PR #92, opened 2024-07-06
added required packages

FrankMittelbach:
some can probably get a better status than "unknown"

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # The LaTeX Tagged PDF repository
 
-See <https://www.latex-project.org/publications/indexbytopic/pdf/>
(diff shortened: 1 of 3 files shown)


u-fischer:
I would put csquotes as "partially-supported", and comment that blockquotes do not work. multicol should be ok (we do have tests with multicol for the para-tagging code.

FrankMittelbach:
> I would put csquotes as "partially-supported", and comment that blockquotes do not work. multicol should be ok (we do have tests with multicol for the para-tagging code.

ok

davidcarlisle (approved):

PR #103, opened 2024-07-09
add amsopn compatibility

mbertucci47:
This lists amsopn as compatible with an included test file. I'm not sure if I should put "phase=III" or not for "supported-through" because tagging doesn't really make sense without setting `testphase`. I've assumed the default is to not list a "supported-through" entry unless one of the other modules (math, table, etc.) are required.

### DIFF
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -91,7 +91,6 @@
  - name: amsopn
    type: package
    status: compatible
-   supported-through: [phase-III]
    issues:
    tests: true
    updated: 2024-07-09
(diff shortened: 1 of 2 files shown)


u-fischer:
math typically needs the math module (remove it in your test file to see the difference in tagging). So it should use `supported-through=[phase-III,math]`. And good math tagging needs mathml files and support in external tools, so the comment from amsmath applies here too.

mbertucci47:
@u-fischer Thanks for the info. Should be updated now to reflect that

FrankMittelbach (approved):

PR #104, opened 2024-07-09
sort order for status

davidcarlisle:
This PR adds a `sorttable_customkey` (just) on `partially-compatible` entries so the `status` column`sorts as compatible,  partially-compatible,  no-support, unknown

testable at

https://davidcarlisle.github.io/tagging-project/tagging-status/


FrankMittelbach (approved):
useful I think (even though one might wonder why it is not sorted alphabetically as the others)

PR #106, opened 2024-07-10
separate KOMA classes

FrankMittelbach:


_data/tagging-status.yml
```diff
@@ -906,3 +899,40 @@
    supported-through: [phase-III,title]   
    updated: 2024-07-05
 
+#------------------------ SSS ----------------------------
+
+ - name: scrartcl
+   ctan-pkg: koma-script
```
davidcarlisle, line 905: We could drop the ctan-pkg key here as most (all?) the Koma script classses have their own ctan catalogue entry eg

https://ctan.org/pkg/scrbook

davidcarlisle (changes requested):

PR #108, opened 2024-07-10
add external issues list

davidcarlisle:
Adds an `external-issues` key that takes a list of URL to  issues not in the tagging-project repositrory, use full URL rather than a github shorthand such as biblatex/1366 so that you could link to gitlab or a mailing list archive or whatever
The shortened form used as link text possibly works best for github issues but is probably usable in all cases.

see biblatex entry at

https://davidcarlisle.github.io/tagging-project/tagging-status


u-fischer:
nice ;-)

FrankMittelbach (approved):
with a comment

PR #109, opened 2024-07-10
add delarray compatibility

mbertucci47:
This lists `delarray` as "compatible". The tagging of
```tex
\begin{array}[t]({c}) 1\\2\\3 \end{array} 
\begin{array}[c]({cr}) 1 & a\\2 & b\\3 & c \end{array} 
\begin{array}[b]({c}) 1\\2\\3 \end{array}
```
with delarray appears to be the same as
```tex
\left(\begin{array}{c} 1\\2\\3 \end{array}\right)
\left(\begin{array}{cr} 1 & a\\2 & b\\3 & c \end{array}\right)
\left(\begin{array}{c} 1\\2\\3 \end{array}\right)
```
without. The PAC checker produces the warning "Possibly inappropriate use of a "Formula" structure element" but this occurs with the non-delarray example above too. Since `array` is already listed as compatible, I assume this is okay.

### DIFF
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -245,9 +245,11 @@
 
  - name: delarray
    type: package
-   status: unknown
+   status: compatible
+   supported-through: [phase-III,math,table]
    issues:
-   updated: 2024-07-06
+   tests: true
+   updated: 2024-07-10
 
  - name: doc
    type: package
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -247,6 +247,8 @@
    type: package
    status: compatible
    supported-through: [phase-III,math,table]
+   comments: "Requires the math module.
+     Use of math tagging requires support in external tools."
    issues:
    tests: true
    updated: 2024-07-10
--- a/tagging-status/testfiles/delarray/delarray-01.tex
+++ b/tagging-status/testfiles/delarray/delarray-01.tex
@@ -1,0 +1,22 @@
+\DocumentMetadata
+  {
+    lang=en-US,
+    pdfversion=2.0,
+    pdfstandard=ua-2,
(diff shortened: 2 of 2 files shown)


u-fischer:
If I got it right it is basically inside the "math blackbox". So I would add the standard math comment, that it needs support in external tools (and mathml etc). But beside this it looks ok.

mbertucci47:
@u-fischer Done

_data/tagging-status.yml
```diff
@@ -245,9 +245,13 @@
 
  - name: delarray
    type: package
-   status: unknown
+   status: compatible
+   supported-through: [phase-III,math,table]
+   comments: "Requires the math module.
+     Use of math tagging requires support in external tools."
```
FrankMittelbach, line 251: As this is not supposed to be a permanent state I suggested we change that line everywhere to

Use of math tagging currently requires ....

And I don't think we need "Requires the math module", this is a leftover from the times we didn't have "supported-through:" Again I think this can be dropped everywhere

@u-fischer @davidcarlisle ?

davidcarlisle: @FrankMittelbach  certainly we don't need to say Requires  the math module,  now we have `supported-through`  adding `currently` OK as well. Alternatively we could remove it all together and just add a global note about `math` support in the paragraph at the top that starts `To use packages or classes together with ...`

Probably simplest to merge this PR as-is, and then we can do the rewording as a global edit over all the package entries separately after the merge?

FrankMittelbach: A global note is simpler (one place) but on the other hand the info is useful if you just consult the ist for a single package, so for now I would leave it  in. And yes, merge and then global edit afterwards

davidcarlisle (approved):

PR #113, opened 2024-07-11
add theorem compatibility

mbertucci47:
This labels the theorem package as "currently-incompatible" and adds a test. Perhaps a comment could be added that the package author @FrankMittelbach does not recommend it.

With that in mind, it would also make sense if "no-support" is more appropriate (feedback welcome).

### DIFF
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -683,10 +683,10 @@
 
  - name: theorem
    type: package
-   status: unknown
+   status: currently-incompatible
    issues:
-   tasks: needs tests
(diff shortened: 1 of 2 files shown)


FrankMittelbach:
it is likely to show up in older documents on arxiv.org and if it isn't too complicated to make it work I would eventually do so for that reason, given that one of the goals is to make old documents accessible (at least to a certain extend) by just recompiling them with a \DocumentMetadata declaration. So currently-incompatible is fine.

FrankMittelbach (approved):

PR #114, opened 2024-07-11
add makeidx compatibility

mbertucci47:
Lists the makeidx package as "compatible" and adds a test file. I'm not sure if indices are supposed to be tagged specially but the structure looks okay to me.

makeidx_struc


FrankMittelbach:
the actual inex entries aren't produced by makeidx, so even if they would be wrong it wouldn't be the fault of this package :-)

FrankMittelbach (approved):

PR #118, opened 2024-07-11
add pstricks compatibility

mbertucci47:
Lists pstricks as "partially-compatible" because it doesn't error but there is no information in the tagging structure. Test file included.

### DIFF
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -605,11 +605,12 @@
 
  - name: pstricks
    type: package
-   status: partially-compatible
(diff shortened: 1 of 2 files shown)


u-fischer:
I would call is similar to tikz `currently-incompatible`. Additionally  a comment should mention that future compatibility will only be with lualatex as the latex-dvips route can not properly support PDF/UA.

mbertucci47:
@u-fischer Done, let me know if anything needs changing

FrankMittelbach (approved):

PR #121, opened 2024-07-11
add showkeys compatibility

mbertucci47:
Lists showkeys as "partially-compatible" because the rules drawn around the labels display as "Path" in the tagging structure. The included test file gives

showkeys_struc

### DIFF
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -646,10 +646,11 @@
 
  - name: showkeys
    type: package
-   status: unknown
+   status: partially-compatible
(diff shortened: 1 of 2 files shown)


u-fischer:
Suggested additional comments:  ...  and the label in the margin is not tagged at all (pdflatex) or as artifact (lualatex). Unproblematic for a draft, but not UA compatible.

mbertucci47:
@u-fischer Additional comment added

FrankMittelbach (approved):

PR #128, opened 2024-07-12
add fancyhdr compatibility

mbertucci47:
Lists fancyhdr as "compatible" and adds a test. My assumption is headers and footers should not be tagged and that's what happens here.


FrankMittelbach (approved):

FrankMittelbach:
With rules being often the case in improper tagging, I think we should verify that this is ok with the rules potentially added by fancyhdr, quite likely they are but ...

PR #130, opened 2024-07-12
add tabulary compatibility

mbertucci47:
Lists tabulary as "compatible", supported with phase-III and table. Test file added; examples from documentation.

### DIFF
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -693,6 +693,14 @@
    issues: [42]
    updated: 2024-07-06
 
+ - name: tabulary
+   type: package
+   status: compatible
+   supported-through: [phase-III,table]
+   issues:
+   tests: true
+   updated: 2024-07-12
+
  - name: tcolorbox
    type: package
    status: partially-compatible
--- a/tagging-status/testfiles/tabulary/tabulary-01.tex
+++ b/tagging-status/testfiles/tabulary/tabulary-01.tex
@@ -36,7 +36,7 @@ a a&b b&c c
 
 \bigskip
 
-With L, R and C columns, and a \verb|\multicolumn|\\
+With L, R and C columns, and a \verb|\multicolumn|
 \begin{tabulary}{\linewidth}{LR|LC}
 1&the rain in spain falls mainly on the plain&
 the rain in spain falls mainly on the plain
--- a/tagging-status/testfiles/tabulary/tabulary-01.tex
+++ b/tagging-status/testfiles/tabulary/tabulary-01.tex
@@ -1,0 +1,52 @@
+\DocumentMetadata
+  {
+    lang=en-US,
+    pdfversion=2.0,
+    pdfstandard=ua-2,
+    testphase={phase-III,math,title,table,firstaid}
+  }
+\documentclass{article}
+\usepackage{tabulary}
+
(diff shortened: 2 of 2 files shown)


FrankMittelbach:
With other table packages \hline and \cline are often causing issues, wonder if those should be added to the test(s)

mbertucci47:
@FrankMittelbach Yes it seems `\hline` works but `\cline` causes issues. I will adjust

davidcarlisle:
@FrankMittelbach  depends how we view it. Unlike say `longtable`, `tabulary` does not touch `\hline`/`\cline` at all so that is just array package code. The package does have (since the release last week) explicit tagging support to guard its trial runs. So not sure if cline issues should mark every table package failing here.

FrankMittelbach:
Well, if we overlooked \cline for array, then we should build some tests to show that this is failing and mark it as partially-compatible (and actually fix it asap)

davidcarlisle:
@mbertucci47  the cline issue you spotted, does the same appear if you just use a tabular with the `array` package?

mbertucci47:
@davidcarlisle Yes, for example
```
\begin{tabular}{crl}
A & B & D \\ \cline{1-2}
E & F & G
\end{tabular}
```
errors

davidcarlisle:
thanks, that also fails without `array`

```
\DocumentMetadata{
  testphase={phase-III,table}
}
\documentclass{article}

\begin{document}

\begin{tabular}{crl}
A & B & D \\ \cline{1-2}
E & F & G
\end{tabular}
\end{document}

```

so I guess we should fix in the latex-lab `table` support rather than flag the individual packages.

mbertucci47:
@davidcarlisle Doesn't the table tagging code automatically load array?

davidcarlisle:
> @davidcarlisle Doesn't the table tagging code automatically load array?

er yes, so strike my "fails without array" comment, (but still means we need to fix this in the generic latex-lab table code not specific for this package)

u-fischer:
looks as if we completly forgot \cline, there is neither code nor testfiles.
Something like this avoid the error. Additionally some tagging code is needed to artifact the leader

~~~~
\DocumentMetadata
  {
    lang=en-US,
    pdfversion=2.0,
    pdfstandard=ua-2,
    uncompress
  }
\documentclass{article}
\makeatletter\ExplSyntaxOn
\def\@cline#1-#2\@nil{%
  \omit
  \@multicnt#1%
  \advance\@multispan\m@ne
  \ifnum\@multicnt=\@ne\@firstofone{&\omit}\fi
  \@multicnt#2%
  \advance\@multicnt-#1%
  \advance\@multispan\@ne
  \leaders\hrule\@height\arrayrulewidth\hfill %artifact tagging needed
  \tbl_gdecr_row_count: %<----------------
  \cr
  \noalign{\vskip-\arrayrulewidth}}
\ExplSyntaxOff
\usepackage{tabulary}

\begin{document}
\begin{tabular}{crl}
A & B & D \\ \cline{1-2}
E & F & G
\end{tabular}
\end{document}
~~~~

davidcarlisle:
@u-fischer  thanks, so we merge this pr as is and handle that elsewhere?

u-fischer:
@davidcarlisle  yes, I just opened an issue for this.

PR #132, opened 2024-07-12
add afterpage compatibility

mbertucci47:
Lists afterpage as compatible and includes a test file. The tagging seems to appear in the correct place

### DIFF
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -52,9 +52,10 @@
 
  - name: afterpage
    type: package
-   status: compatible
+   status: partially-compatible
    issues:
    tests: true
+   tasks: more tests needed
    updated: 2024-07-12
 
  - name: alltt
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -52,10 +52,10 @@
 
  - name: afterpage
    type: package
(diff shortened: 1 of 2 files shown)


davidcarlisle:
I agree that the supplied test file looks OK, but given what `afterpage` does I'd be astonished if it manages to be compatible here in general (since I have not done anything to make it compatible).  In the test file there is nothing much happening at the end of page so the afterpage insertion happens at a relatively safe place, but in general the insertion can happen within an arbitrary structure, say mid list, or between the lines of a math `align`  or (my personal favourite between the first and second `$` in `$$`  (so after math mode detected but before a display math starts)  It seems unlikely that inserting in these places does not mess up the tag nesting, but I'll try to make an example.

davidcarlisle:
I think the test is fine and shows it doesn't completely break, but I think `partially-compatible` with a comment more tests needed would be a safer status than claiming compatibility.

mbertucci47:
@davidcarlisle Suggested changes added

FrankMittelbach:
I was about to say the same, one should really take a look through usages of afterpage in TL and build matching use cases. However, it isn't that unlikely that it works out of the box in most cases if not in all.

PR #133, opened 2024-07-12
add amscd compatibility

mbertucci47:
Lists amscd as compatible with the usual math disclaimer and adds a test file. The content in the tag structure in acrobat is mostly nonsense but that's also true for other math objects.

### DIFF
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -75,10 +75,12 @@
 
  - name: amscd
    type: package
-   status: unknown
+   status: compatible
+   supported-through: [phase-III,math]
+   comments: "Use of math tagging currently requires support from external tools."
    issues:
-   tasks: needs tests
-   updated: 2024-07-06
+   tests: true
+   updated: 2024-07-12
 
  - name: amsmath
    type: package
--- a/tagging-status/testfiles/amscd/amscd-01.tex
+++ b/tagging-status/testfiles/amscd/amscd-01.tex
@@ -1,0 +1,39 @@
+\DocumentMetadata
+  {
+    lang=en-US,
+    pdfversion=2.0,
+    pdfstandard=ua-2,
+    testphase={phase-III,math,title,table,firstaid}
+  }
+\documentclass{article}
+\usepackage{amscd,amsmath}
+
+\DeclareMathOperator{\End}{End}
+\DeclareMathOperator{\cov}{cov}
+\DeclareMathOperator{\cf}{cf}
+\DeclareMathOperator{\add}{add}
+\DeclareMathOperator{\non}{non}
+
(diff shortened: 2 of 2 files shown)


FrankMittelbach:
> but that's also true for other math objects

for which "other" math objects is this the case?

mbertucci47:
@FrankMittelbach All I mean is the plain text tag labels are not particularly useful for math. For example if you add
```tex
\begin{equation}
\prod_{x\in |X|}\frac{1}{1-t^{\deg x}}
\end{equation}
```
to the test file the tag structure looks like

math_struc

FrankMittelbach:
ok, so what you are saying (I think) is that if math is not having an associated file or imbedded mathml in the inner part of the structure is essentialy useless --- which it is, and what this remark means about the need for external tools.

However, I wonder if we should in the basic version (where the formula content is described via alt text) mark the actual formula content as artifact. Wdyt @u-fischer @davidcarlisle ?

davidcarlisle:
@FrankMittelbach 

> However, I wonder if we should in the basic version (where the formula content is described via alt text) mark the actual formula content as artifact. Wdyt @u-fischer @davidcarlisle ?

artifact seems wrong (since it is the main content)  but would it be better when there is no mathml to give the tex source as ActualText instead of, or as well as, Alt? rather than just letting the raw character data fall through to the tagging?
I did try to do that just now but failed to get the right tagpdfsetup incantation @u-fischer ?

FrankMittelbach:
> artifact seems wrong (since it is the main content) but would it be better when there is no mathml to give the tex source as ActualText instead of, or as well as, Alt? rather than just letting the raw character data fall through to the tagging?
> I did try to do that just now but failed to get the right tagpdfsetup incantation @u-fischer ?

I suggested artifact, because it might otherwise be read  even if there is alt + Actual text, I think some  viewers / AT tools do that. But I agree in principle just ActualText should be fine (maybe plus alt). Guess it need experimenting with different tools and/or ask around. But somehow we should suppress the strange data currently shown. @u-fischer ?

but perhaps we should not discuss this here further but open an issue for this as it isn't urgent and this pr should get merged.

u-fischer:
I don't think that ActualText is correct unless we start to add tagging inside the Formula and want to handle single symbols. Artifact doesn't sound right either. The best way imho to get rid of the "strange data" is to use unicode-math plus some better definitions for \frac etc.

FrankMittelbach:
what's wrong (though not good) with ActualText containing the TeX source?

u-fischer:
Well the TeX source is not the "actual text". Beside this it wouldn't help: you would still see the "strange data", but on the other side you would affect copy&paste.

PR #136, opened 2024-07-12
add bm compatibility

mbertucci47:
Lists bm as compatible and adds a test file. Compatibility code is contained in latex-lab-mathpkg

### DIFF
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -171,10 +171,12 @@
 
  - name: bm
    type: package
-   status: unknown
+   status: compatible
+   supported-through: [phase-III,math]
+   comments: "Use of math tagging currently requires support from external tools."
    issues:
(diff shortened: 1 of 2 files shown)


FrankMittelbach:
Can't remember if bm uses "poor man's bold" in some cases but if it does I would expect that to fail tagging. So perhaps that needs testing an at least a further comment (if it fails indeed)

davidcarlisle:
@FrankMittelbach  the poor man's bold code is redefined in `latex-lab-mathpkg.dtx`  

```
% \subsection{bm}
% Similar to the amsbsy code in latex-lab-amsmath for \cs{pmb} we need to mark up 
% copies as artifact:
% 
```

FrankMittelbach:
so much the better, then I see no reason to not merge this.

PR #140, opened 2024-07-13
add axessibility

u-fischer:


_data/tagging-status.yml
```diff
@@ -133,6 +133,12 @@
    issues:
    updated: 2024-07-06
 
+ - name: axessibility
+   type: package
+   status: no-support
+   comments: loads tagpdf but it not compatible with the current version. It patches
```
davidcarlisle, line 139: > loads tagpdf but it not compatible with the current version. It patches
loads tagpdf but it is not compatible with the current version. Its patches

_data/tagging-status.yml
```diff
@@ -133,6 +133,12 @@
    issues:
    updated: 2024-07-06
 
+ - name: axessibility
+   type: package
+   status: no-support
+   comments: loads tagpdf but it is not compatible with the current version. Its patches 
+             clash with the math module. phase-III+math should be used instead.
```
FrankMittelbach, line 140: suggestion:

... which provides broader tagging support without the need to load an additional package.

FrankMittelbach:
I handle that in oe of my commits instead.

PR #142, opened 2024-07-13
TLC3 recommended packages (letter A)

FrankMittelbach:


FrankMittelbach:
perhaps a few more could me marked compatible (just by the look of it)

davidcarlisle (approved):
I'll merge, some may need ctan-pkg adding (unless you checked this already)  but that can be added later

davidcarlisle:
actually these:

```
amssymb
antpolt
anttor
apalike
authordate1-4
avant
```

davidcarlisle:
ctan-pkg added at 874590a

PR #143, opened 2024-07-14
clarify what the number mean, it is easy to overlook that one set is …

FrankMittelbach:
…across all tables and the one below just for packages


davidcarlisle (approved):

davidcarlisle:
ctan-pkg added at 2ed675b

PR #146, opened 2024-07-14
move scrletter and change type

mbertucci47:
Moves scrletter to the packages section and changes type to "package" since it is a sty file.


FrankMittelbach (approved):
good point, thanks

PR #149, opened 2024-07-14
devanagari not devnag package

davidcarlisle:
devnag is the preprocessor name not a package or class: changed entry to devanagari

### DIFF
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -1109,7 +1109,7 @@
    tests: true
    updated: 2024-07-10
 
- - name: devnag
+ - name: devanagari
    type: package
    status: unknown
    issues:


FrankMittelbach (commented):
I wonder if we shouldn't just drop them. The reference in TLC3 is just to old 8-bit engine support which for such languages is at best suboptimal. So I doubt it will ever get a status anyway. Same for bengali then. Ref is TLC3 II-341

davidcarlisle:
@FrankMittelbach  OK wth me, I can delete this PR and you could drop that at same time as your next "letter" update

FrankMittelbach:
Ok I do it WTH the next batch after you fix my mistakes in I to M

davidcarlisle:
package entry will be dropped

PR #190, opened 2024-07-17
add common packages to tagging-status.yml

mbertucci47:
Adds several commonly-used packages and classes to tagging-status.yml without commenting on their status. If any are too obscure to be included please let me know.

### DIFF
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -382,6 +382,14 @@
    tasks: needs tests
    updated: 2024-07-13
 
+ - name: attachfile2
+   type: package
+   status: unknown
+   issues:
+   tests: false
+   tasks: needs tests
+   updated: 2024-07-17
+
  - name: atveryend
    type: package
    status: compatible
@@ -449,6 +457,14 @@
    tests: false
    tasks: needs tests
    updated: 2024-07-14
+
+ - name: backref
+   type: package
+   status: unknown
+   issues:
+   tests: false
+   tasks: needs tests
+   updated: 2024-07-17
 
  - name: baskervillef
    type: package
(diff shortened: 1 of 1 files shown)


car222222:
I think that this list should be pretty much comprehensive: with nothing being excluded as simply because it is "too obscure".

FrankMittelbach:
> I think that this list should be pretty much comprehensive: with nothing being excluded simply because it is "too obscure".

I do a have a different opinion on this. While I agree that it is good to have an overview on the status of the whole universe of LaTeX packages with respect to tagging support, it is equally important (and in my opinion even more so) to have a curated set of packages, that cover most requirements that documents can have and do this with (reasonable) mature implementations. This set of packages is the the group that should be covered first because one has to balance available resources with a project result that brings a lot of benefits in most situation.

Of course, any such group is a biased selection, so it shouldn't be taken as something that can't be augmented. However, if we add every package and  end up with 5000+ CTAN packages out of which initially more than 4500 have no known status and only a few of them are then likely to get one, then it obsources both what is necessary to address as well as to actually schedule work in a sensible way.

This is why I added and focused on *just* the packages covered by TLC3 (i.e. around), because TLC3 is the result of evaluating *all* packages in CTAN not that long ago(biased for sure) for importance, coverage, and quality in the respective space.
In my opinion that (plus a few more ) is the set the project should focus on initially and what we should also strongly suggest as a good basis for usage by users.  This should be crosschecked and augmented by, say ,the usage found across arXiv.org (recent 5 years say).

This is why I'm not sure that it is a good idea to simply start adding further packages if there is not an explicit request for that package, just because it happens to be on CTAN.

mbertucci47:
@FrankMittelbach I fully agree that it is not desirable to include every possible CTAN package. I've tried to make sure the packages I've added to the list in #214 are all ones I've seen "in the wild", i.e. in questions/answers on TeX-SX or in my own experience in other people's real documents (mostly math documents). But I recognize your points and I think #216 is a good solution.

PR #200, opened 2024-07-18
add tests for comment package

PetrSojka:
It seems OK - commented parts were omitted from the PDF and became artifacts.


tagging-status/comment/comment-01.tex
```diff
@@ -0,0 +1,33 @@
+DocumentMetadata
```
mbertucci47, line 1: Missing a `\`

FrankMittelbach (changes requested):
please add missing \

PR #201, opened 2024-07-18
add tests for fixme (only basic use)

wiobber:


_data/tagging-status.yml
```diff
@@ -1533,11 +1533,11 @@
 
  - name: fixme
    type: package
-   status: unknown
+   status: partially-compatible
```
FrankMittelbach, line 1536: looking at the pdf structure generated I think this can (for now) be classified as compatible

FrankMittelbach (changes requested):
suggest to classify as compatible

_data/tagging-status.yml
wiobber: stats updated to 'compatible'

mbertucci47:
I think the changes to tagging-status.yml got lost somewhere

davidcarlisle:
@mbertucci47  that's odd I just looked at the commit history and not sure quite where that got dropped. so it should be `compatible` ? I can fix locally  but I wanted to check I understood the comment

davidcarlisle:
@mbertucci47  0292cc1

PR #204, opened 2024-07-18
add csvsimple compatibility

mbertucci47:
Lists csvsimple as compatible and adds a test file. Comments that `\csvautotabularray` is not yet supported because tabularray is not compatible

### DIFF
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -1001,11 +1001,13 @@
 
  - name: csvsimple
    type: package
-   status: unknown
+   status: compatible
+   comments: "`\\csvautotabularray` etc. are not supported since tabularray is
+             currently incompatible."
+   supported-through: [phase-III,table]
    issues:
-   tests: false
-   tasks: needs tests
-   updated: 2024-07-14
+   tests: true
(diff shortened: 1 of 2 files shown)


_data/tagging-status.yml
```diff
@@ -1001,11 +1001,13 @@
 
  - name: csvsimple
    type: package
-   status: unknown
+   status: compatible
+   comments: "`\\csvautotabularray` etc. are not supported since tabularray is
```
FrankMittelbach, line 1005: given the comment (and not knowing how important the command (and the etc) is for using the package, I wonder if it sholdb't be better classified as partially-compatible

FrankMittelbach (changes requested):
use partially-compatible

mbertucci47:
@FrankMittelbach Changed as requested. My thought was that this is just a tacked-on extra feature of the package (the analogous and more probably more commonly-used commands are `\csvautotabular`, `\csvautolongtable`, and `\csvautobooktabular` which use kernel tabular, longtable, and tabular with booktabs, respectively). But I understand why you might want total compatibility to call it "compatible"

PR #212, opened 2024-07-19
Add extra status for maintainers activly want to be involved/are working on it

TeXhackse:
As discussed during the developers workshop before TUG I'd love to have some kind of status indicating that testfiles are provided elsewhere and the package maintainers are actually working on it, including the link to the repo. 

To show an example I would – in case you approve this PR - edit the entry for ragged2e in the following way:

```yaml
- name: ragged2e
  type: package
  status: in-progress
  tests: true
  source-repository: https://gitlab.com/TeXhackse/ragged2e
  supported-through: phase-III
  updated: 2024-07-19
```
edit: move the repo to a separate key as suggested in the comments.

I tried to find out how the website is built, which I did not manage on first try, so I just tried to find out how it works. Idea was to help, in case it's not helpful just drop this PR. Not sure if you want to stick to the 3 letter naming for the columns. I tried to… 

Not sure about the naming … but I thought `ìn-progress` is a good idea, but that makes the 3 letter thing a bit more complicated. Currently I added `xpm` where `m` stands for `maintained` which would also be an option… but that usually would not indicate that someone is actually working on this feature extension.

### DIFF
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -25,6 +25,7 @@
 # -----                  -------
 # compatible             This package or class works without any issues when tagging is enabled
 # partially-compatible   The package or class is currently partially compatible, see comments column
+# in-progress		 Status can be choosen by maintainers to mark they are aware and already working on it. This means testfiles should not be included in th tagging-project but in the package/class repository.
 # currently-incompatible The package or class is currently incompatible with the tagging code, but we expect it to be updated eventually. 
 # no-support             This package or class or class is incompatible with the tagging code and we do *not* believe that it will ever be supported.
 # unknown                The status of this package or class is not known
--- a/tagging-status/index.md
+++ b/tagging-status/index.md
@@ -42,6 +42,7 @@ If you encounter a problem with a package or class for which there is no issue i
 {% assign xpackages = site.data.tagging-status | where: "type", "package" %}
 {% assign xpc = xpackages | where: "status", "compatible" %}
 {% assign xpp = xpackages | where: "status", "partially-compatible" %}
+{% assign xpm = xpackages | where: "status", "in-progress" %}
 {% assign xpi = xpackages | where: "status", "currently-incompatible" %}
 {% assign xpn = xpackages | where: "status", "no-support" %}
 {% assign xpu = xpackages | where: "status", "unknown" %}
@@ -49,11 +50,12 @@ If you encounter a problem with a package or class for which there is no issue i
 Statistics: The table currently holds **{{xpackages | size }}** packages out of which
 **{{xpc| size }}**  are fully `compatible`,
 **{{xpp | size }}** `partially-compatible`,
+**{{xpm | size}}**`in-progress`,
 **{{xpi | size }}** `currently-incompatible`, and
 **{{xpn | size }}** have  `no-support`.
 The status for the remaining **{{xpu | size }}** packages is `unknown`.
 
-{% assign packages = xpc | concat: xpp | concat: xpi | concat: xpn | concat: xpu %}
+{% assign packages = xpc | concat: xpp | concat: xpm | concat: xpi | concat: xpn | concat: xpu %}
 
 {% include_relative status-table.md %}
 
@@ -64,6 +66,7 @@ The status for the remaining **{{xpu | size }}** packages is `unknown`.
 {% assign xpackages = site.data.tagging-status | where: "type", "class" %}
 {% assign xpc = xpackages | where: "status", "compatible" %}
 {% assign xpp = xpackages | where: "status", "partially-compatible" %}
+{% assign xpm = xpackages | where: "status", "in-progress" %}
 {% assign xpi = xpackages | where: "status", "currently-incompatible" %}
 {% assign xpn = xpackages | where: "status", "no-support" %}
 {% assign xpu = xpackages | where: "status", "unknown" %}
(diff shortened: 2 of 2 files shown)


davidcarlisle:
don't worry about the "three letter" liquid variables they are just used in one place (actually two as the code isn't currently shared between classes and packages) to grab the filtered lists for iteration and have the counts available. They are just internal names just to get round some restrictions in filtering lists in liquid (you can filter on values of the yaml keys but not so easily on boolean combinations of those keys, so I filter separately and then concat the results)

I suppose main thing to decide is if the repo link should be part of the free text comment field as you show or if we extend the set of keys. We already have `external-issues: ["https://github.com/plk/biblatex/issues/1366"]`  to link to issues outside the tagging repo, so we could have `source-repository:  https://gitlab.com/TeXhackse/ragged2e` or some such to be just used for repositories where tagging adjustments are happening not general package sources. It would take a bit more liquid templating  to make a link out of the supplied URL  compared to a markdown link in the comments which would "just work" already but might give a more structured data?

davidcarlisle:
It may not be clear enough but I Ithink the existing `supported-through: package`  key is supposed to indicate that fixes have been (or are being) made in the package so that, along with a status of `compatible` /`partialy-compatible`/`currently-incompatible` and a new link to the external repo may be enough and not need `in-progress` adding as a new status?

TeXhackse:
The `source-repository` suggestion is fine for me and I can change that. 

The extra status is thing I think which might be important for end-users and also concerning the question who is in charge. 

It makes a huge difference if users report to you and that extra status would make them first blame me and if I realize it's not something I can do about it then I can report back to you. So the idea was to shift a bit of traffic as well of work to maintainer repos instead of yours. 

Only having the repo would not indiciate in any way that there is anyone actively working on it.

davidcarlisle:
as I say that was the intention of `supported-through: package` to indicate that it was being updated at the package rather than patched in firstaid by the latex project,  It may not be clear enough but we probably don't want both as that's likely to be confusing.  currently only hyperref has this entry which was intended to indicate, as you say, that hyperref tagging issues should be tracked in the hyperref repo not here.

> This only indicates which latex-lab code is required for the support, doesn't it?
> But it doesn't resolve the question of who is in charge.

latex-lab modules can be listed but the special value `package` was intended to mean that tagging was being maintained by the package author (although we don't currently give a link to the repository, adding a link would be an improvement)

TeXhackse:
I see, then yes, it's not clear and was described differently yesterday. Thanks for adding that info.

davidcarlisle:
@TeXhackse  we were of course making up the yaml format as we went along adding/changing as needed and witht many examples originally, also I had to miss the conference, so missed any discussion at the workshop but I think we can do something.  see what any of the others think...

davidcarlisle:
I would suggest that we keep the current `supported-through` key meaning taking a list of testphase modules plus the special value `package` to mean package with author support, then add a new key

`external-repository: .....`

which can be used in addition to give a source repository link. In the html display **package** would be a html link to that repository if this is used.
The existing `external-issues` field could also be used to link to specific issues in addition to this top level repository link.

@TeXhackse  does that sound OK?

davidcarlisle:
@TeXhackse  the entry is now

```
 - name: ragged2e
   type: package
   status: partially-compatible
   priority: 2
   supported-through: [phase-III,package]
   package-repository: "https://gitlab.com/TeXhackse/ragged2e"
   external-issues: "https://gitlab.com/TeXhackse/ragged2e/-/issues/9"
   issues:
   tests: false
   tasks: needs tests
   updated: 2024-07-15
```

the  `package`  entry in the `supported-through` field now becomes a link to the repo as there is a `package-repository:` field.  in addition I added an `external-issues`  entry which as before generates a link to the gitlab issue.

Is that enough to cover the feature request here?

Theer needs to be more documentation at the top on what the fields mean, and this specific entry  can change the `tests: false` and `tasks:needs tests`  but  This is rendering now in the `ragged2e` entry at  https://latex3.github.io/tagging-project/tagging-status/

TeXhackse:
Sorry, I didn't see your question before. It's okay. that way. I can live with that. Actually most important is, that we all know how its meant to be read. Thanks a lot!

davidcarlisle:
@TeXhackse  OK thanks I'll close here then.
I note in the gitlab issue in your repo you comment you are waiting for feedback from  thae latex team but I wasn't sure what "both implementations" meant?

TeXhackse:
Ah @davidcarlisle that's about #213.

Thing is I have 2 different implementations as one is the one relying on the latex-lab block code. Its not in the repo yet, as I had too many things to do this week (and finish the proceedings paper) but it's working just needs doc. 

The default one should stay the default one of course. So I need to switch between those implementations based on if the latex-lab code is loaded (or available somehow else) or not and would like to have some recommended way to switch, so that not every package is doing their own thing once again.

davidcarlisle:
@TeXhackse  ah Ok it's logged here good, thanks for the link.

FrankMittelbach:
> @TeXhackse the entry is now
> 
> ```
>  - name: ragged2e
>    type: package
>    status: partially-compatible
>    priority: 2
>    supported-through: [phase-III,package]
>    package-repository: "https://gitlab.com/TeXhackse/ragged2e"
>    external-issues: "https://gitlab.com/TeXhackse/ragged2e/-/issues/9"
>    issues:
>    tests: false
>    tasks: needs tests
>    updated: 2024-07-15
> ```
> 
> the `package` entry in the `supported-through` field now becomes a link to the repo as there is a `package-repository:` field. in addition I added an `external-issues` entry which as before generates a link to the gitlab issue.
> 
> Is that enough to cover the feature request here?
> 
> There needs to be more documentation at the top on what the fields mean, and this specific entry can change the `tests: false` and `tasks:needs tests` but this is rendering now in the `ragged2e` entry at https://latex3.github.io/tagging-project/tagging-status/

@davidcarlisle as far as I can see these special keys aren't documented at the top of the yaml file (or elsewhere), I think we better do that so that we know what to add in such cases. Can you please do that?

FrankMittelbach:
ah saw too late that this was a merge request ... nothing to merge here only to update the source ... sorry

PR #211, opened 2024-07-19
add floatpag and update lineno.

u-fischer:

PR #214, opened 2024-07-20
Add many packages

mbertucci47:
Taking @car222222's [comment](https://github.com/latex3/tagging-project/pull/190#issuecomment-2235251497) to heart...

Adds many packages to tagging-status.yml without commenting on status. Many like the font packages could probably be marked as compatible without test files but I'll save that for another PR. I tried to include only packages that I think have a nonzero number of users and are not
1. very specific to a journal, conference, or university
2. obviously reliant on tikz or pstricks since these are not yet compatible anyways

This is in no way complete, just the result of a cursory browse through CTAN.

floatpag was removed from the addition since I see it will be added in #211.

### DIFF
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -2365,14 +2365,6 @@
    tasks: needs tests
    updated: 2024-07-18
 
- - name: floatpag
-   type: package
-   status: unknown
-   issues:
-   tests: false
-   tasks: needs tests
-   updated: 2024-07-18
-
  - name: floatrow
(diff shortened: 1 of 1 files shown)


davidcarlisle:
hmmm I guess so, I suspect there are diminishing returns in making the list much bigger, certainly the flat yaml file structure and basic liquid filtering wasn't really set up to manage a full ctan database, but looking at the additions here by eye I notice several "well known"  packages, let's go with it and see how the structure holds up...

PR #216, opened 2024-07-21
priority

davidcarlisle:
This adds a `priority` field to entries, only used if status is `unknown`  it should be an integer between 1 and 9  inclusive and `unknown` entries  are only  listed in the generated table if they have priority 1 or 2.

handling this proved trickier than expected as the `where_expr;` filter (despite being documented as available since jekyll 3.2) doesn't appear to work on gh-pages, which seems to claim it is jekyll 3.9.5 but anyway I think it's working now filtering "by hand" in a for loop.

The resulting display can be seen at

https://davidcarlisle.github.io/tagging-project/tagging-status/

which now shows 547 rather than 951 entries, although all the data is still in the YAML.

See discussion comments from me and Frank at

https://github.com/latex3/tagging-project/pull/214#issuecomment-2241338408

https://github.com/latex3/tagging-project/pull/190#issuecomment-2241548182

The initial assignment of priorities is a bit arbitrary for ease of bulk editing, all unknown entries with an updated date up to 2024-07-15 get priority 2 and all later ones get 9, this ensures that the TLC3 list gets shown as do all later additions with a state other than unknown. Individual packages could be adjusted after the merge.

### DIFF
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -9,6 +9,7 @@
 #  type              required   string          "class" or "package" (may need eg tikz or beamer themes in the end?)
 #  status            required   string          Status of package, see list below.
 #  ctan-pkg          optional   string          Name for ctan.org/pkg/?? link if different to name
+#  priority          optional   integer         Priority of work for unknown entries (omitted from table if > 3)
 #  comments          optional   markdown-string Free text markdown comments
 #  references        optional   integer-list    List of integers referencing the bibliography in references.yml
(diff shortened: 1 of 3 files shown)


davidcarlisle:
@FrankMittelbach  adjusted as you requested (elsewhere)

* entries with priority < 5 now show
* priority applies to `currently-incompatible` and `uknown`
* priority value is displayed (currently as a subscript to the status)
* still the case that current priorities are all either 2 or 9

https://davidcarlisle.github.io/tagging-project/tagging-status/#classes

PR #217, opened 2024-07-21
add topcapt compatibility

mbertucci47:
Lists [topcapt](https://www.ctan.org/pkg/topcapt) as compatible and adds a test file.

### DIFF
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -6692,11 +6692,10 @@
 
  - name: topcapt
    type: package
-   status: unknown
+   status: compatible
    issues:
-   tests: false
-   tasks: needs tests
-   updated: 2024-07-18
+   tests: true
+   updated: 2024-07-21
 
  - name: totalcount
    type: package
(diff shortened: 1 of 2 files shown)


davidcarlisle:
Not sure this should be `compatible` it doesn't error but the caption is essentially just text I think, the extensive \caption support in

latex2e/required/latex-lab/latex-lab-float.dtx

isn't picked up as `\topcaption` is just a modified copy of the original caption code, it doesn't actually use any of the redefined caption code.

% \subsection{Handling captions}

mbertucci47:
In Acrobat it seems to be tagged the same as with `\caption`. Here's the tags from the example:

topcapt_struc

The first is with `\topcaption`, the second with `\caption`. I think `\topcaption` is just using an unmodified `\@caption` so still picks up the tagging code

davidcarlisle:
oh I must have missed something sorry, I'll re-check

davidcarlisle:
@mbertucci47  I misread the package code...

PR #218, opened 2024-07-21
add hypcap compatibility

mbertucci47:
Lists [hypcap](https://www.ctan.org/pkg/hypcap) as "currently-incompatible" and adds two test files. The first test shows that hypcap is unnecessary since the tagging code achieves its purpose by moving the anchor to the top of floats. The second test shows it errors if given one of the package options like `figure` that redefines the float environment.

It's possible "no-support" makes more sense.

### DIFF
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -3070,11 +3070,13 @@
 
  - name: hypcap
    type: package
-   status: unknown
+   status: currently-incompatible
+   comments: "The tagging code already moves anchors to the top of floats
+              so this package is unnecessary. Errors if given option
+              `figure`, `table`, etc."
    issues:
-   tests: false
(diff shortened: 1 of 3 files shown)


davidcarlisle:
@mbertucci47  this and other "currently incompatible" PR will show merge conflicts as I've merged the priority change so unknown` and `currently-incompatible` entries have `priority`.  If you wish to adjust the that's fine but otherwise there is no need: we can fix up the conflict in each case while merging to main, and the conflicts are just "expected adjustments" and don't affect any discussion on the PR.  (I picked this PR at random, this comment could be made on all of them)

mbertucci47:
@davidcarlisle Whatever is easiest. If you can do them without manual adjustments to each PR then let's go with that

PR #219, opened 2024-07-21
mark font packages as compatible

mbertucci47:
This marks several packages that do nothing but set up fonts as "compatible" without providing test files, though I did a cursory check with each one. I can add tests if needed. If no more packages should be added to the list please feel free to close, but I figured most of these would get hidden by #216 so no harm in listing their compatibility in the plaintext version.

### DIFF
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -346,11 +346,10 @@
 
  - name: andika
    type: package
-   status: unknown
+   status: compatible
    issues:
    tests: false
-   tasks: needs tests
-   updated: 2024-07-18
+   updated: 2024-07-21
 
  - name: animate
    type: package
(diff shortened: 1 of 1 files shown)


FrankMittelbach:
Anything that can be marked compatible even if way out of a core set is helpful imho and the prio setting will not prevent it from being displayed. It is just that I don't think it is good to display a lot of packages as unknown of the likelihood that they get looked at in the near future is close to nil. Using prio for them helps scheduling any work nd not polluting the tables with packages most people will never use.

FrankMittelbach:
I'll look at this or when get to a decent computer but I guess it'll be fine

davidcarlisle:
@FrankMittelbach  I checked a couple of packages and structurally the PR is Ok modulo expected merge conflicts with the priority branch, so merging

PR #222, opened 2024-07-21
add footnpag compatibility

mbertucci47:
Lists [footnpag](https://www.ctan.org/pkg/footnpag) as "partially-compatible" because the footnote tagging looks correct but it redefines `\maketitle` in a way that breaks footnotes in `\title`. Test files included.


davidcarlisle:
merging, but note I left in the `priority: 9` even though currently priority is unused for `partially-compatible` as noted in a different PR it may be best to filter priority for all status except `compatible`

PR #223, opened 2024-07-21
add footnoterange compatibility

mbertucci47:
Lists [footnoterange](https://www.ctan.org/pkg/footnoterange) as "partially-compatible" since the footnote label "i-j" is incorrectly tagged as normal text and the hyperref support is broken since it relies on the existence of an `Hfootnote` counter which is not provided when the tagging code is loaded.

### DIFF
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -2839,7 +2839,7 @@
 
  - name: footnoterange
    type: package
-   status: partially-compatible
+   status: currently-incompatible
    priority: 2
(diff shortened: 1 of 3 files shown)


davidcarlisle:
judgement call, but your description sounds closer to currently-incompatible?  (just a question not saying you should definitely change)

mbertucci47:
Hmm yes it does. I guess I was thinking the first problem is similar to that of endnotes, which @FrankMittelbach said to call partially-compatible, and not working with hyperref did not push it into fully incompatible. Since hyperref is more important for tagged pdfs I'll change it to currently-incompatible.

PR #231, opened 2024-07-22
add autoaligne compatibility

mbertucci47:
Lists [autoaligne](https://www.ctan.org/pkg/autoaligne) as "partially-compatible" and supported by "phase-III,math". It is only partially compatible because the vertical alignment types `h` and `b` produce a `! Missing $ inserted.` error. Looking at the package code, I think this is because of a use of `\vtop` and `\vbox` in a way the math tagging code does not like. Test files with working and failing examples included.

### DIFF
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -531,11 +531,13 @@
 
  - name: autoaligne
    type: package
-   status: unknown
+   status: partially-compatible
+   supported-through: [phase-III,math]
+   comments: "Errors with vertical alignment types `h` and `b`.
+              Use of math tagging currently requires support from external tools."
    issues:
-   tests: false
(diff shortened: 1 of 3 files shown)


davidcarlisle:
This was/is  priority 9 so as it stands it wouldn't be listed if  `unknown` or `currently-incompatible`  but would be listed if this PR merged and it becme `partially-compatible` . Perhaps that is  OK, or should we use `priority` to filter everything except `compatible` which might be simpler to document anyway? 

Adding Frank as reviewer to comment although this PR could be merged anyway, the change in filtering would be in the liquid templating in the markdown file so not affecting this PR directly.

davidcarlisle:
priority filtering for partially-compatible has been added elsewhere so the above comment no longer applies

PR #232, opened 2024-07-22
add cancel compatibility

mbertucci47:
Lists [cancel](https://www.ctan.org/pkg/cancel) as "compatible" with the math comment and adds a test file.

### DIFF
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -1005,11 +1005,12 @@
 
  - name: cancel
    type: package
-   status: unknown
+   status: compatible
+   supported-through: [phase-III,math]
+   comments: "Use of math tagging currently requires support from external tools."
    issues:
-   tests: false
-   tasks: needs tests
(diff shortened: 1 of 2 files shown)


davidcarlisle:
I'm not sure this should be marked as compatible, even if we defer the math use to a better mathml generator to be applied, the text use of `\xcancel{blub}`  gets tagged as text `)))blub`  which I guess is sub-optimal (not sure how a strikeout should be tagged) @u-fischer ?

u-fischer:
I agree with @davidcarlisle. I'm not sure how the math/mathml should look like, but the text version should imho be tagged with a Span + a TextDecorationType attribute with value LineThrough.

davidcarlisle:
so I think this needs to be `currently-incompatible` as even if it doesn't error the tagging is actively misleading as it loses the fact that the text is crossed out

mbertucci47:
Status has been updated with a comment

Issue #139, opened 2024-07-12
mathtools' `mathic` option incompatible with math grabbing code


mbertucci47:
The [mathtools](https://www.ctan.org/pkg/mathtools) option `mathic` errors with `testphase={phase-III,math}`. This makes sense as it redefines `\(...\)`.
```tex
\DocumentMetadata
  {
    lang=en-US,
    pdfversion=2.0,
    pdfstandard=ua-2,
    testphase={phase-III,math}
  }
\documentclass{article}
\usepackage{mathtools}

\begin{document}

\itshape
Compare these lines: \par
\mathtoolsset{mathic} % or \mathtoolsset{mathic=true}
Subset of \(V\) and subset of \(A\). \par
\mathtoolsset{mathic=false}
Subset of \(V\) and subset of \(A\).

\end{document}
```
errors with
```
Runaway argument?
\MH_fi: V\) and subset of \(A\). \par \mathtoolsset {mathic=false} Su\ETC.
! File ended while scanning use of \__math_grab_dollar:w.
<inserted text> 
                \par
```

FrankMittelbach:
My expectation would be that mathtools would largely work with a number of such things failing, so that we end up with partially-supported and a list of items (envs or options) that do not work. What is therefore needed is determining which parts are failing (and which work), i.e. more tests.

mbertucci47:
I don't claim to understand everything the mathic code is doing but the following naive changes seem to work for this simple document.
```tex
\DocumentMetadata
  {
    lang=en-US,
    pdfversion=2.0,
    pdfstandard=ua-2,
    tagging=on,
  }
\documentclass{article}
\usepackage{mathtools}
%\usepackage{unicode-math}

\MHInternalSyntaxOn
\def\MT_mathic_true: {
  \MH_if_boolean:nF {math_italic_corr}{
    \MH_set_boolean_T:n {math_italic_corr}
    \MH_if_boolean:nTF {robustify}{
      \MH_let:NwN \MT_mathic_redeffer: \DeclareRobustCommand
    }{
      \MH_let:NwN \MT_mathic_redeffer: \renewcommand
    }
    \MH_let:NwN \MT_begin_inlinemath: \(
    %\renewcommand*\({
    \MT_mathic_redeffer:*\({
      \relax\ifmmode\@badmath\else
      \ifhmode
        \MH_if_dim:w \fontdimen\@ne\font>\z@
          \MH_if_dim:w \lastskip>\z@
[...]

PR #245, opened 2024-07-22
add mathtools compatibility

mbertucci47:
Lists [mathtools](https://www.ctan.org/pkg/mathtools) as "partially-compatible" and adds several test files, each corresponding to a subsection of the documentation. Currently only mathtools-07.tex errors and is logged in #139. For the rest the tagging looks okay but of course it should be looked at closely by someone more knowledgeable than me.

### DIFF
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -4129,9 +4129,12 @@
 
  - name: mathtools
    type: package
-   status: unknown
-   tasks: needs tests
-   updated: 2024-07-06
+   status: partially-compatible
+   supported-through: [phase-III,math]
+   comments: "`mathic` key produces errors. Use of math tagging currently requires support from external tools."
+   issues: [139]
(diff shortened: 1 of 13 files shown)


car222222:
Did this fix mathtools so that it is now fully compatible?

mbertucci47:
@car222222 Are you referring to this commit: https://github.com/latex3/mathtools/commit/22cca27e33668c2f66d0d33bafb5110334f4a9f3 ?

FrankMittelbach:
@car222222 we consider amsmath as only partially compatible (for a number of reasons, eg because of the math support not finished and because we haven't tested several aspects). Given that mathtools is based on amsmath it is therefore also at best partially compatible which is what is currently stated. As for forther tests: we are certainly grateful  for any that are  being prepared and checked ... don't think that that has happened so far --- fancy writing some and checking them with ngpdf?

PR #253, opened 2024-07-23
remove duplicate entry

mbertucci47:
Removes the duplicate entry for [CooperHewitt](https://www.ctan.org/pkg/cooperhewitt). The actual file name is `CooperHewitt.sty` (with caps).


davidcarlisle:
d.@*n! I saw that duplicate appear while resolving the merge conflict then forgot to clean up.  Thanks:-)

Issue #16, opened 2023-10-11
\includegraphics[height=..] produces "Use of \??? doesn't match its definition."


John02139:
I'm seeing an error when I use \includegraphics with the option [height = ..], but not with [width = ].  Example below.

```
\DocumentMetadata{ 
	testphase = {phase-III},
}
\documentclass[12pt]{report} 
\usepackage{lipsum,graphicx,mwe} % mwe shown only because it's the source of these sample images

\begin{document}
\chapter{One}
\lipsum[5]
\bigskip

\includegraphics[alt={text}, height=3.75cm]{example-image-b}\quad\includegraphics[alt={text}, width=5cm]{example-image-b}
% using [height=] produces a LaTeX error, but using [width=] does not
\end{document}
```
leads to this error
```
! Use of \??? doesn't match its definition.
<argument> \???  
                 ! LaTeX Error: Premature end in fp expression.
l.12 ... = {text}, height=3.75cm]{example-image-b}
                                                  \quad\includegraphics[alt ...
? 
```
I am running pdfTeX with:
    LaTeX2e <2023-06-01> patch level 1
    L3 programming layer <2023-08-29>

u-fischer:
Thanks for the report. Yes, there is a bug in the code for the graphic tagging. It assume that the \Gin@scalex is a number, which isn't always true.

josephwright:
Something like
```latex
   \fp_set:Nn \l_@@_graphic_scale_fp
    {
      \str_compare:VnTF \Gin@scalex { ! }
        { \Gin@scaley }
        { \Gin@scalex }
    }
```
We could use `\tl_if_head_eq_charcode:nNTF` I guess, as we only need to know about the first char, but I suspect the `str` test is clearer.

John02139:
This is resolved in the 2023/11/01 release.

hpvd:
`comments` and `related issues` for `graphicx` in https://latex3.github.io/tagging-project/tagging-status/
could be updated according to the solution of this issue...

PR #263, opened 2024-07-24
change sillypage to compatible

mbertucci47:
Compatibility added with [sillypage \#16](https://github.com/cereda/sillypage/pull/16). It's missing a priority entry but I'll let someone else decide how important this silly package is :-)


u-fischer:
As it has been mentioned in the LaTeX companion it is clearly very important!

PR #269, opened 2024-07-25
add mparhack compatibility

mbertucci47:
Lists [mparhack](https://www.ctan.org/pkg/mparhack) as "compatible" and adds a test file. Whatever mparhack is changing behind the scenes doesn't seem to affect the tagging.

### DIFF
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -4769,7 +4769,8 @@
 
  - name: mparhack
    type: package
-   status: compatible
+   status: partially-compatible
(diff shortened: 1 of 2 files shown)


u-fischer:
I would call it only partially compatible. It will error if you use the debug option (`\usepackage[debug]{mparhack}`). Also as it patches core kernel commands (`\@outputpage`, `\@outputdblcol`, `\@addmarginpar`) it is a package that should be replaced/disabled long term.

mbertucci47:
Status updated

PR #282, opened 2024-07-25
add adforn and adfsymbols compat

mbertucci47:
Lists adforn, adfarrows, and adfbullets as "partially-compatible" since the symbols are missing ToUnicode values.


cfr42:
how do I figure out what the unicode values should be? is there an equivalent of the purely-decorative declaration for images? (this would not always be right, but I don't know what else `adforn` might use?

PR #286, opened 2024-07-25
add lua-ul compatibility

mbertucci47:
Lists lua-ul as "currently-incompatible" for the reasons discussed in #232. Unlike with cancel, the text is tagged as-is, however there should be TextDecoration attributes to indicate strike-through, underline, and highlight.

(By the way, where can you see attributes for a Span? Is it possible without opening up the uncompressed pdf file?)

### DIFF
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -4251,12 +4251,12 @@
 
  - name: lua-ul
    type: package
-   status: unknown
+   status: currently-incompatible
    priority: 2
+   comments: "Missing TextDecoration attributes."
    issues:
-   tests: false
-   tasks: needs tests
-   updated: 2024-07-15
+   tests: true
+   updated: 2024-07-25
 
  - name: lua-widow-control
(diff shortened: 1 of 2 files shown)


davidcarlisle:
acrobat pro can show attributes (although I don't have that these days)  or the online ngpdf viewer, for example

u-fischer:
It depends. If the attribute is attached to a structure, you can find it if go to the properties of a structure element and then to attributes or attribute classes. E.g. here for the center. But it is also possible to attach attributes directly to a MC and then you have to look into the properties of the container..

~~~~
\DocumentMetadata{uncompress,pdfversion=2.0,testphase=phase-III}
\documentclass{article}
\begin{document}

\begin{center}
abc
\end{center}

\ExplSyntaxOn
aa\pdf_bdc:nn{Span}{/C /center}aaa\pdf_emc: aaa
\ExplSyntaxOff
\end{document}
~~~~

image

PR #296, opened 2024-07-25
add parskip compatibility

mbertucci47:
Lists parskip as compatible and adds a test file.

### DIFF
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -5605,7 +5605,6 @@
  - name: parskip
    type: package
    status: compatible
-   priority: 9
    issues:
    tests: true
    updated: 2024-07-25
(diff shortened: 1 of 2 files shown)


FrankMittelbach (changes requested):
Why was this made prio 9 at some point?  This is a TLC3 package so no prio or prio 1 please

davidcarlisle:
@FrankMittelbach  that may have been me, adding the initial priorities basically ran off update date, so if a packge was updated after the main addition of tlc3 it may have been wrongly priotised. priority is ignored for `compatible` packages in any case so it can be deleted.

mbertucci47:
Wasn't me :) Fixed

FrankMittelbach:
> Wasn't me :) Fixed

:-) didn't meant to indicate that

PR #312, opened 2024-07-26
add uri compatibility

mbertucci47:
Lists [uri](https://www.ctan.org/pkg/uri) as partially-compatible because if hyperref is loaded, there are two empty tags added after every link. This does not occur with just `\url` or `\href`, for example. Test included (pulled from tlc3-examples).

### DIFF
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -7615,12 +7615,12 @@
 
  - name: uri
    type: package
-   status: unknown
+   status: partially-compatible
    priority: 2
    issues:
-   tests: false
-   tasks: needs tests
-   updated: 2024-07-15
+   comments: "Produces empty tags after links if hyperref loaded."
+   tests: true
+   updated: 2024-07-26
 
  - name: url
    type: package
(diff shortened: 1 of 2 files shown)


FrankMittelbach (approved):

u-fischer:
hm I don't see which empty tags you mean. The structure looks fine here.

mbertucci47:
@u-fischer Here's what a portion of the structure tree looks like for me, with lualatex:

uri_struc

mbertucci47:
More minimally:
```tex
\DocumentMetadata
  {
    lang=en-US,
    pdfversion=2.0,
    pdfstandard=ua-2,
  }
\documentclass{article}
\usepackage{hyperref}
\usepackage{uri}

\title{uri tagging test}

\begin{document}

\url{https://latex-project.org} % to compare

\href{https://latex-project.org}{random text} % to compare

\doi{10.1111/coin.12165}

\asin{0201362996}
\end{document}
```
uri_struc2

u-fischer:
Ah. That is not serious. Empty containers appear now and then. A comment that one should perhaps look at some time if one can avoid them is enough.

PR #327, opened 2024-07-26
add xy compatibility

mbertucci47:
Lists [xy](https://www.ctan.org/pkg/xypic) as partially-compatible because in text mode, no tagging occurs for an `\xymatrix`. Test included. Perhaps currently-incompatible is better.


FrankMittelbach:
I think I would go with incompatible for now

davidcarlisle:
> I think I would go with incompatible for now

I did that while handling the existing merge conflict

PR #353, opened 2024-07-29
More math packages

mbertucci47:
Lists letterswitharrows, luamathalign, mathalpha, mathdots, mattens, old-arrows, oubraces, scalerel, stackrel, subdepth, upgreek, and yhmath as compatible and adds test files.

Lists steinmetz as partially-compatible because it needs to tag the picture used to draw the math symbol as Artifact.

Lists subeqn as currently-incompatible because it is incompatible with amsmath, as it defines a `subequations` environment.

Also fixes a typo palantino->palatino.

### DIFF
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -4736,13 +4736,14 @@
 
  - name: letterswitharrows
    type: package
-   status: unknown
+   status: compatible
    included-in:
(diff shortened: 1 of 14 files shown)


davidcarlisle:
> Also fixes a typo palantino->palatino.
oops thanks:-)

PR #360, opened 2024-07-30
add metalogo and metalogox compatibility

mbertucci47:
Lists [metalogo](https://www.ctan.org/pkg/metalogo) and [metalogox](https://www.ctan.org/pkg/metalogox) as compatible and adds tests.

### DIFF
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -5569,19 +5569,21 @@
 
  - name: metalogo
    type: package
-   status: compatible
+   status: partially-compatible
    included-in:
    priority: 9
    issues:
+   comments: Logos need ActualText.
    tests: true
    updated: 2024-07-30
 
  - name: metalogox
    type: package
-   status: compatible
+   status: partially-compatible
    included-in:
    priority: 9
    issues:
+   comments: Logos need ActualText.
    tests: true
    updated: 2024-07-30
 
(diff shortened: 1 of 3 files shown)


FrankMittelbach (commented):
I might be mistaken, but I think those have to be partially compatible, because, logos like LaTeX or LuaTeX etc should really be tagged with some actualtext to take out the weirdness of the normal TeX related logoes.

By the way do we do that yet for the basic logoes?

mbertucci47:
Ah I was assuming the behavior of the default `\LaTeX` etc. was correct. Will fix

u-fischer:
partially compatible. They inherit in part from our tagging of TeX and LaTeX, but e.g. `\LaTeXe` copy and paste as `LaTeXLATEX2ε` which is clearly wrong.

mbertucci47:
There's a few odd things with copy/paste. If you copy just one line from the following document you get something reasonable, but if you copy both lines you get
```
LaTeX
LaTeX
LATEX2ε
```
```tex
\DocumentMetadata
  {
    lang=en-US,
    pdfversion=2.0,
    pdfstandard=ua-2,
  }
\documentclass{article}
\begin{document}
\LaTeX

\LaTeXe
\end{document}
```
If they're both `\LaTeXe` you get
```
LATEX2ε
LaTeX
LATEX2ε
```
That's with luatex. With pdftex I get the same thing but with no line breaks or spaces.

PR #388, opened 2024-07-31
add fnpos compatibility

mbertucci47:
Lists [fnpos](https://www.ctan.org/pkg/fnpos) as compatible and adds a test. The package redefines `\@makecol` which seems dangerous but I couldn't get it to break or mess up the tagging for multicol or twocolumn.

### DIFF
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -3298,11 +3298,13 @@
 
  - name: fnpos
    type: package
-   status: compatible
+   status: partially-compatible
    included-in:
    priority: 9
    issues:
+   comments: Redefines `\\@makecol`.
    tests: true
+   tasks: Revisit once OR is altered.
    updated: 2024-07-31
 
  - name: fnumprint
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -3298,13 +3298,12 @@
 
  - name: fnpos
    type: package
-   status: unknown
+   status: compatible
    included-in:
    priority: 9
    issues:
-   tests: false
-   tasks: needs tests
-   updated: 2024-07-18
+   tests: true
+   updated: 2024-07-31
 
  - name: fnumprint
    type: package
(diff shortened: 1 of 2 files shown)


FrankMittelbach:
I would expect it to break, because the tagging for footnote also modifies the OR. and one or the other would overwrite each other or so I would have thought. But I see that your tests involve footnotes. Hmmm ... :-)

davidcarlisle:
@FrankMittelbach  presumably explicit use of the new sockets such as `@makecol/outputbox` would not go weill id `\@makecol` got redefined?

FrankMittelbach:
it probably wouldn't :-). But anyway, packages like that will have to change to not use patching but sockets and hooks at some point so even if it now works for some reason it will stop working eventually for a bit. Not sure how to capture that ... if we want to capture it.

FrankMittelbach:
maybe  by  calling it  partially-compatible with a tasks not to revisit when OR is altered

mbertucci47:
Comment and tasks added

u-fischer:
With pdftex it breaks tagging as the marks needed to close the MC at the page break are lost. E.g.

~~~~
\DocumentMetadata
  {
    lang=en-US,
    pdfversion=2.0,
    pdfstandard=ua-2,
    uncompress,xmp=false
  }
\documentclass{article} 

\usepackage{fnpos}
\usepackage[nopar]{kantlipsum}
\textheight=5cm
\begin{document}

\kant[1-2]
\end{document}
~~~~

gives in the first page stream something like this

~~~~
/text <</MCID 0 >> BDC
BT
/F28 9.9626 Tf 148.712 657.235 Td [(As)]
.....
/F28 9.9626 Tf 43.989 0 Td [(since)]TJ/F39 9.9626 Tf( )Tj/F28 9.9626 Tf 24.753 0 Td [(kno)28(wledge)]TJ
ET %<--------------here emc is missing.
/Artifact BMC
BT
/F28 9.9626 Tf 303.133 495.578 Td [(1)]TJ
ET
EMC
~~~~

mbertucci47:
@u-fischer So should this be currently-incompatible then?

FrankMittelbach:
i would think so

PR #389, opened 2024-07-31
add fnpct compatibility

mbertucci47:
Lists [fnpct](https://www.ctan.org/pkg/fnpct) as compatible with a note that the `ranges` option does not work with hyperref.


FrankMittelbach:
would lean towards partly-compatible then

mbertucci47:
Status updated

PR #395, opened 2024-07-31
add amsrefs compatibility

mbertucci47:
Lists amsrefs as compatible and adds a few tests.

### DIFF
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -453,13 +453,12 @@
 
  - name: amsrefs
    type: package
-   status: unknown
+   status: compatible
    included-in: [tlc3, arxiv01]
    priority: 2
    issues:
-   tests: false
-   tasks: needs tests
-   updated: 2024-07-13
+   tests: true
+   updated: 2024-07-31
 
  - name: amssymb
    ctan-pkg: amsfonts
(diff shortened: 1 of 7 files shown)


davidcarlisle:
the references get tagged as a list but shouldn't specific bibliography tagging get used @u-fischer @FrankMittelbach ?

mbertucci47:
@davidcarlisle I haven't looked into the code but a simple example doesn't show any bibliography specific tagging.
```tex
\DocumentMetadata
  {
    lang=en-US,
    pdfversion=2.0,
    pdfstandard=ua-2,
  }
\documentclass{article}

\begin{document}

Some text and a citation: \cite{inbook-full}

\bibliographystyle{amsalpha}
\bibliography{xampl}

\end{document}
```

davidcarlisle:
I'll leave this one for the others to review but note there is additional bib code in testphase={bib}, (I think)

u-fischer:
@davidcarlisle there is currently no special code for the bibliography - it is still open if there is a need to use Bibentry somewhere or if a list as we use now is enough. The main point of the bib code is to ensure that there is a reference from the citation to the bib.

davidcarlisle:
ah OK, so I'll merge thanks

FrankMittelbach:
@davidcarlisle @u-fischer I stll wonder if we shouldn't at least add in cases like this a `tasks=review when ...` because if there is a decision to make them up differently the package would need an update.

davidcarlisle:
@FrankMittelbach  yes I added a comment in `tasks`

PR #398, opened 2024-08-01
add bibstyles

mbertucci47:
Several of the packages listed in the yaml file are bibliography styles with accompanying 209 style files (or in the case of [nar](https://ctan.org/pkg/nar), no .sty file). This lists those as compatible without adding test files, though I did a quick check with each that they didn't break tagging. Three of them are not distributed with texlive so a note was added for those.

The listed packages are apalike, [astron](https://ctan.org/pkg/astron), [chicago](https://ctan.org/pkg/chicago), [jmb](https://www.ctan.org/pkg/jmb), [named](https://ctan.org/pkg/named), [nar](https://ctan.org/pkg/nar), and [newapa](https://ctan.org/pkg/newapa).

I also removed [urlbst](https://ctan.org/pkg/urlbst) since it's not a package but a perl script.

### DIFF
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -562,13 +562,12 @@
  - name: apalike
    ctan-pkg: bibtex
    type: package
-   status: unknown
+   status: compatible
    included-in: [tlc3]
    priority: 2
    issues:
    tests: false
-   tasks: needs tests
-   updated: 2024-07-13
+   updated: 2024-07-31
 
  - name: arabi
    type: package
@@ -664,13 +663,13 @@
 
  - name: astron
    type: package
-   status: unknown
+   status: compatible
    included-in: [tlc3]
    priority: 2
    issues:
    tests: false
-   tasks: needs tests
-   updated: 2024-07-13
+   comments: Not distributed with texlive.
+   updated: 2024-07-31
 
  - name: asymptote
    type: package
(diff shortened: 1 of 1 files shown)


davidcarlisle:
@mbertucci47 thanks for catching this, I'll wait for @FrankMittelbach  and @u-fischer  to comment before merging this.

I think the .sty are potentially for latex2e as well as 2.09 but in general they typically overwrite `\cite`  and/or the bibliography list formatting, so they may be impacted by any updates to core bibliography code. 

For `amsrefs` I added

```
   tasks: "review after final decision on tagging for bibliography lists"
```

at Frank's suggestion I wonder if we should do something similar for these or even more formally distinguish them with a yaml boolean

```
bibliography-package: true
```

or something..

Frank and Ulrike have thought more about this than me,  so leaving for review.

FrankMittelbach:
It might be worth adding a key like `package-type` with values like `bibliography`, `index`,  `output-routine`, `footnote` and perhaps some others over time, because any of those are likely needing a revisit if we move hooks into core LaTeX in that space. On the other hand it might be over-engineering that yaml file without much gain ... dunno

u-fischer:
well amsrefs is rather special. For with the current tagging code the rule of thumb for "normal" cite/bib packages is if they work with hyperref tagging works too (as similar adaptions are used). So I'm not sure if they need a special marker (but I didn't check all the packages).

davidcarlisle:
OK thanks, let's merge then, we may need to periodically adjust the status anyway as things change.....

PR #407, opened 2024-08-02
add enotez compatibility

mbertucci47:
Lists enotez as partially-compatible because the `\endnote`s are not tagged as `<Note>`s and parent-child warnings are produced if `\endnotemark` contains a `\ref` command (see second test).

### DIFF
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -2727,13 +2727,14 @@
 
  - name: enotez
    type: package
-   status: unknown
+   status: partially-compatible
    included-in: [tlc3]
    priority: 2
    issues:
-   tests: false
-   tasks: needs tests
-   updated: 2024-07-14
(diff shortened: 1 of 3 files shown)


FrankMittelbach (commented):
wonder if that should be labeled currently-incompatible. @u-fischer thoughts?

u-fischer:
I think partially-compatible. It looks more or less sensible if hyperref is loaded, as then the link creates a reference. Without hyperref it is not ok. 

The parent-child warning in the second test can be avoided with `\endnotemark[\ref*{en:test}]`  (that is the typical link in link problem). 

Independant from tagging, enotez doesn't handle destinations correctly. You get warnings about duplicated destinations and if you try to reference the third endnote the link is wrong, I opened an issue about that https://github.com/cgnieder/enotez/issues/51

PR #417, opened 2024-08-02
add newunicodechar compatibility

mbertucci47:
Lists [newunicodechar](https://www.ctan.org/pkg/newunicodechar) as compatible and adds a test.

### DIFF
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -6441,13 +6441,12 @@
 
  - name: newunicodechar
    type: package
-   status: unknown
+   status: compatible
    included-in: [arxiv001]
    priority: 7
    issues:
-   tests: false
-   tasks: needs tests
-   updated: 2024-07-18
+   tests: true
+   updated: 2024-08-02
 
  - name: newverbs
    type: package
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -6444,7 +6444,7 @@
    status: partially-compatible
    included-in: [arxiv001]
    priority: 7
-   issues:
+   issues: [417]
    tests: true
    comments: Needs a way to support ActualText for characters.
    updated: 2024-08-02
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -6441,11 +6441,12 @@
 
  - name: newunicodechar
    type: package
-   status: compatible
+   status: partially-compatible
    included-in: [arxiv001]
    priority: 7
    issues:
    tests: true
+   comments: Needs a way to support ActualText for characters.
    updated: 2024-08-02
 
  - name: newverbs
--- a/tagging-status/testfiles/newunicodechar/newunicodechar-01.tex
+++ b/tagging-status/testfiles/newunicodechar/newunicodechar-01.tex
@@ -1,0 +1,26 @@
+\DocumentMetadata
+  {
+    lang=en-US,
+    pdfversion=2.0,
+    pdfstandard=ua-2,
+    testphase={phase-III,math,table,title,firstaid}
+  }
+\documentclass{article}
+\usepackage{newunicodechar}
+%\usepackage[T1]{fontenc} % for pdflatex
+
+\newunicodechar{♢}{<>}
+\newunicodechar{⧫}{!}
+\newunicodechar{⬠}{(P)}
+\newunicodechar{μ}{\mu}
+
+\begin{document}
+♢
+
+⧫
+
+⬠
+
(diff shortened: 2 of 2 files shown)


davidcarlisle:
Hmm this package is often used to fake missing characters from the font but here the tagging is only seeing the replacement not the orginal character. Eg the diamond operator `♢` in the supplied test file is just expanding to `<>` but (sometimes at least) that's just a font fallback issue and the construct ought to have actual text preserving the `♢`.

the package though is quite general and just makes the character active and could have any definition not necessarily a "faked character" so arguably it is up to the package user to tag the replacement definition in an appropriate way,

Adding Ulrike for review before merging.

mbertucci47:
> arguably it is up to the package user to tag the replacement definition in an appropriate way

That's what I was thinking

davidcarlisle:
in the end the command probably should have an optional argument defaulting to the original character to supply an actual text,  rather than trust users to use lower level tagging commands within the replacement text. But that may come under "possible future enhancements" not really a compatibilty issue with current code. Not sure. See what the others say.

FrankMittelbach:
I would call it partially compatible and refer in the comments to the issue of needing a concept to support actual text linking to this discussion. Perhaps we need an issue category of "partially-compatible" nut just "currently incompatible" there to keep track of longer discussions.

mbertucci47:
Status updated

_data/tagging-status.yml
```diff
@@ -6441,13 +6441,13 @@
 
  - name: newunicodechar
    type: package
-   status: unknown
+   status: partially-compatible
    included-in: [arxiv001]
    priority: 7
    issues:
```
FrankMittelbach, line 6447: explicitly refer to [417] I would say for more details
```
issues: [417]
```

FrankMittelbach (changes requested):

u-fischer:
Why should tagging care about the original input character? If someone wants to add "hello duck" to their document like this
`\newunicodechar{🦆}{hello duck}🦆`  then there is no need to retain the emoji. The intended text is hello duck, and that is there. I mean we are not trying to retain other commands that create output either.

FrankMittelbach (approved):

mbertucci47:
That was my original thought but I think @FrankMittelbach is saying the package should natively offer an option to replace at the pdf level, for the cases where it is a truly "faked" character like `\newunicodechar{♢}{$\diamond$}`

FrankMittelbach:
> That was my original thought but I think @FrankMittelbach is saying the package should natively offer an option to replace at the pdf level, for the cases where it is a truly "faked" character like `\newunicodechar{♢}{$\diamond$}`

wasn't me saying that was @davidcarlisle but yes, basically I think the package should offer a way to give actual text and why not also with an option to easily give the unicode char. For example, in theaboe it shouldn't end up showing a math formula.

u-fischer:
I think that there is a general need for a command that can be used to give an arbitrary symbol an unicode meaning so `\ToUnicode{symbol}{unicode}`. That is also needed for packages like marvosym etc. But I don't quite see why newunicodechar should need here a special interface.

FrankMittelbach:
I think the difference is that this package offers a user way to declare unicode characters. In comparison packages like marvosym set up a fixed set of unicode characters. So these packages need an interface to be able to easily say, this symol represents such and such unicode char. Of course that interface could also be use in addition to `\newunicodechar`. But there it makes some sense to think that it could be `\newunicodechar[actual=...]{....}{...}` and in the absense use the unicode value by default. At least that is what I thought @davidcarlisle  was suggesting.

u-fischer:
> and in the absense use the unicode value by default. 

Why that? If someone creates an active char to print (P) or \alpha in their document, one should assume that they want (P) and \alpha as unicode value. In most cases the replacement code of a \newunicodechar is fine "unicode wise": `\newunicodechar{🦆}{hello duck}` works well. `hello duck` is the intended meaning and if we leave it alone it is better accessible and behaves better as with an actualtext. `\newunicodechar{🦆}{\tikz{\duck}}` on the other side *is* a problem. But `\tikz{\duck}` alone is a problem too. And imho in both case we need `\tikz[actualtext=duck]{\duck}`.

davidcarlisle:
@u-fischer  I agree that if you view it as just an active character with an arbitrary definition then it's the responsibility of the user to tag the replacement text (or not) but when I see this package used on stackexchange I'd say the majority use is not that, it's coping with font deficiencies  to typeset what is logically a single character. But OK I'll merge:-)

PR #419, opened 2024-08-02
add dashrule compatibility

mbertucci47:
Lists [dashrule](https://ctan.org/pkg/dashrule) as compatible and adds a test. I'm not sure about this one because the rule commands appear as "Path" or worse "PathPathPath..." in tags, however this is also the behavior of `\rule`. Maybe these should be Artifacts? But then the same should be true for `\rule`.

### DIFF
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -2266,13 +2266,13 @@
 
  - name: dashrule
    type: package
-   status: unknown
+   status: compatible
    included-in: [tlc3]
    priority: 2
    issues:
-   tests: false
-   tasks: needs tests
-   updated: 2024-07-14
+   tests: true
+   comments: "`\\hdashrule` commands appear as Path in tags but so does `\\rule`."
+   updated: 2024-08-02
 
  - name: dashundergaps
    type: package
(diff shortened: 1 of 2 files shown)


u-fischer:
rules should normally be artifacts unless they have some semantic meaning, in which case they should be a span with actual text.

mbertucci47:
Okay, I will change the status then. Does this mean the tagging for `\rule` should be changed too? I encounter similar issues with "PathPath..." a lot with things like `\fbox{text}` too where I don't know what the correct tagging should be.

mbertucci47:
Status changed to partially-compatible

FrankMittelbach:
> Okay, I will change the status then. Does this mean the tagging for `\rule` should be changed too? I encounter similar issues with "PathPath..." a lot with things like `\fbox{text}` too where I don't know what the correct tagging should be.

I think one way to say this is that `\fbox` in core LaTeX is not handled correctly and should perhaps get its own issue.

FrankMittelbach (approved):

u-fischer:
>  Does this mean the tagging for \rule should be changed too? 

Yes. It would naturally be easier if one could simply leave them alone, but in the end we will have to review various uses of rules (also in tabulars etc).

mbertucci47:
See #420

PR #423, opened 2024-08-02
add bookmark compatibility

mbertucci47:
Lists bookmark as compatible and adds a test.

Not sure if there's anything special I should be looking for here. The [pdfa best practices](https://pdfa.org/wp-content/uploads/2019/06/TaggedPDFBestPracticeGuideSyntax.pdf) don't say much about bookmarks other than it's recommended the pdf opens to the bookmark page by default, and the `lang` must be set. The former seems true (but up to the pdf viewer ultimately) and the latter is warned about if not set in `\DocumentMetadata`.

For the example document, taken from the bookmark doc, the PAC checker complains that it skips from `\section` to `\subsubsection`, but that seems like user error.

### DIFF
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -1297,13 +1297,12 @@
 
  - name: bookmark
    type: package
-   status: unknown
+   status: compatible
    included-in: [tlc3, arxiv01]
    priority: 2
    issues:
-   tests: false
-   tasks: needs tests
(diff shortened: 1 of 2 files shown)


u-fischer:
bookmark is okay. The main point where it can fail is to have actions to destinations without  structure destinations but that has been handled some time ago.

PR #427, opened 2024-08-02
arxiv status and easy pkgs

mbertucci47:
This adds `arxiv001` to packages I missed before, but shouldn't change the priority since they all had `[tlc3]` to begin with. Adds `[arxiv1]` to llncs and changes priority.

Lists bbold, bigints, calligra, calrsfs, dsfont, esvect, and mathbbol as compatible with tests.

Lists auxhook, environ, everyshi, and trimspaces without tests. 

Lists [ascii](https://www.ctan.org/pkg/ascii-font) and [ifsym](https://www.ctan.org/pkg/ifsym) as partially-compatible because some symbols need ToUnicode/ActualText values.

Changes "droidmono" to "droidsansmono".

### DIFF
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -115,7 +115,7 @@
  - name: abraces
    type: package
    status: compatible
-   included-in: [tlc3]
+   included-in: [tlc3, arxiv001]
    priority: 2
    supported-through: [phase-III,math]
    comments: "Use of math tagging currently requires support from external tools."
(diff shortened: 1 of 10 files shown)


FrankMittelbach:
> Changes "droidmono" to "droidsansmono".

drats ... your earned yourself an entry into the tlc3 lottery (well half one because you didn't submit an errata :-) ) 

That package was called draidmono at one point, sigh

PR #442, opened 2024-08-03
add packages

mbertucci47:
Lists bigfoot and citation-style-language as currently incompatible with tests and links to issues.

Lists needspace as compatible with a test.

Lists pgfrcs as compatible without a test (this is just a utility sub-package of pgf).

### DIFF
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -1153,12 +1153,13 @@
 
  - name: bigfoot
    type: package
-   status: unknown
+   status: currently-incompatible
    included-in: [tlc3, arxiv001]
    priority: 2
+   issues: [410]
    comments: To be supported or not?
-   tasks: needs tests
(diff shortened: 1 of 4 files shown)


u-fischer:
hm not sure what to do with pgfrcs. It is not really fully compatible: that is the part that loads the code for the pdf management of pgf, and we have patches in the pdfmanagement to make their resource management compatible with ours. So there is an open task here, but it won't error.

mbertucci47:
@u-fischer I see. I can mark it partially-compatible or if it's easier just remove it from the PR and leave as unknown

FrankMittelbach:
> @u-fischer I see. I can mark it partially-compatible or if it's easier just remove it from the PR and leave as unknown

either or would work I guess

davidcarlisle:
I'll merge then make it pgfrcs partially-compatible

PR #462, opened 2024-08-04
add undolabl compat

mbertucci47:
Lists [undolbl](https://www.ctan.org/pkg/undolabl) as compatible and adds a test.

This is a dependency of [pageslts](https://www.ctan.org/pkg/pageslts) which is in tlc3 but based on some recent TeX-SX posts I think this package is broken with current latex.

### DIFF
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -9492,6 +9492,15 @@
    tests: true
    updated: 2024-07-26
 
+ - name: undolabl
+   type: package
+   status: compatible
+   included-in:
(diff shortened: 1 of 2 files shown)


mbertucci47:
@FrankMittelbach Sorry, what I wrote is confusing. pageslts is incompatible, but undolbl is compatible. Another PR will deal with pageslts

FrankMittelbach (approved):

FrankMittelbach:
ah I see now how I should have read it :-)

PR #483, opened 2024-08-05
add eso-pic compatibility

mbertucci47:
Lists [eso-pic](https://www.ctan.org/pkg/eso-pic) as partially-compatible because the content of `\AddToShipoutPictureBG` is not always tagged as Artifact, for example when it contains an image. I'm not actually sure about this but in my mind if the content appears on every page then it doesn't make sense in the flow of the document.


u-fischer:
eso-pic is nowadays a small wrapper round the shipout/backgroud and shipout/foreground hook. So the task imho here is to handle the tagging of these hooks. Probably suspending tagging by default plus some tools to reenable if wanted is needed here.  Something like this (needs some testing to be sure that the literals are not nested somewhere). 

~~~~
\DocumentMetadata
  {
    lang=en-US,
    pdfversion=2.0,
    pdfstandard=ua-2,
    uncompress,xmp=false
  }
\documentclass{article}
\usepackage{graphicx,lipsum}
\ExplSyntaxOn
\cs_set:Npn \__shipout_picture_overlay:n #1 {
    \__shipout_init_page_origins:
    \kern -\c__shipout_horigin_tl \scan_stop:
    \vbox_to_zero:n {
      \kern -\c__shipout_vorigin_tl \scan_stop:
      \unitlength 1pt \scan_stop:
      \hbox_set_to_wd:Nnn \l__shipout_tmp_box \c_zero_dim
        {\tagmcbegin{artifact} %with pdftex \pdf_bmc:n{Artifact}/pdf_emc: is more straightforward, but 
        % with luatex we should set the attribute to avoid nested bmc. 
         \SuspendTagging{shipout}\ignorespaces #1 \hss
         \ResumeTagging{shipout}\tagmcend
        }
      \box_set_ht:Nn \l__shipout_tmp_box \c_zero_dim
      \box_set_dp:Nn \l__shipout_tmp_box \c_zero_dim
      \box_use:N \l__shipout_tmp_box
      \tex_vss:D
    }
}
\ExplSyntaxOff
\begin{document}
\ExplSyntaxOn
xxx\AddToHook{shipout/background}{\put(3cm,-5cm)
   {TEXT%
    \ResumeTagging{shipout}
    \tagmcend\includegraphics{example-image-duck}
    \tagmcbegin{artifact}\SuspendTagging{shipout}}}
blub
\ExplSyntaxOff

\AddToHook{shipout/background}{\put(3cm,-5cm){MORE TEXT}}

blub

\end{document}
~~~~

FrankMittelbach:
Guess this should get an issue with a link to  Ulrike's comments

PR #498, opened 2024-08-06
add flushend compatibility

mbertucci47:
Lists [flushend](https://www.ctan.org/pkg/flushend) as partially-compatible and adds tests. Partial since it errors if any content is added with `\atColsEnd`.

### DIFF
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -3508,13 +3508,13 @@
 
  - name: flushend
    type: package
-   status: unknown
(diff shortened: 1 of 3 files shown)


FrankMittelbach:
Have you checked out any of the other features? I wonder how important that particular feature is but if there are also others that break tagging perhaps it should be labeled currently-incompatible. Dunno

FrankMittelbach (approved):
then lets call it partially ...

PR #509, opened 2024-08-06
add algorithm and algorithmic compatibility

mbertucci47:
Lists algorithm as currently-incompatible because it defines the `algorithm` environment with the float package so is not tagged as a float.

Lists algorithmic as compatible and adds a test. The `algorithmic` environment is implemented as a list which I guess is okay for tagging algorithms.

Removes the "algorithms" entry since that's just the CTAN name for these two packages.

### DIFF
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -294,37 +294,28 @@
  - name: algorithm
    ctan-pkg: algorithms
    type: package
-   status: unknown
+   status: currently-incompatible
    included-in: [arxiv5]
    priority: 4
    issues:
-   tests: false
-   updated: 2024-07-28
+   tests: true
+   comments: "`algorithm` environments not tagged as floats. See float package."
+   updated: 2024-08-06
 
  - name: algorithmic
    ctan-pkg: algorithms
    type: package
-   status: unknown
+   status: compatible
    included-in: [arxiv5]
    priority: 4
    issues:
-   tests: false
-   updated: 2024-07-28
+   tests: true
+   updated: 2024-08-06
 
  - name: algorithmicx
    type: package
    status: unknown
    included-in: [tlc3, arxiv5]
-   priority: 2
-   issues:
-   tests: false
-   tasks: needs tests
-   updated: 2024-07-14
-
- - name: algorithms
-   type: package
-   status: unknown
-   included-in: [tlc3]
    priority: 2
    issues:
    tests: false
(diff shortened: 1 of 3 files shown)


davidcarlisle:
> Removes the "algorithms" entry since that's just the CTAN name for these two packages.

@FrankMittelbach  TLC3 index errata?

FrankMittelbach:
> > Removes the "algorithms" entry since that's just the CTAN name for these two packages.
> 
> @FrankMittelbach TLC3 index errata?

that's an odd one. it is true the bundle on ctan generates 2 .sty files but if you want to get to the documentation then that is called "algorithms", e.g. texdoc algorithms finds it but using algorithm gives you a list of options. hmmm

_data/tagging-status.yml
```diff
@@ -294,22 +294,23 @@
  - name: algorithm
    ctan-pkg: algorithms
    type: package
-   status: unknown
+   status: currently-incompatible
    included-in: [arxiv5]
```
FrankMittelbach, line 298: with errata for TLC3 this should then say [TLC3,arxiv5].  The prio can stay at 4 as it is only a mentioning in TLC

FrankMittelbach (changes requested):
add TLC3 ref for algorithm

_data/tagging-status.yml
mbertucci47: Done. Should it be added to algorithmic too?

FrankMittelbach (approved):

_data/tagging-status.yml
FrankMittelbach: my understanding is that algorithmic is really a sub package that normally not used on its own (though it can). Anyway, in TLC it is only a remark on page I-322 and I have now changed there from algorithms to algorithm, so not  it should not get TLC3 in included-in

mbertucci47: Neither one loads the other and they can be used separately. algorithm just defines the float `algorithm`, algorithmic defines the `algorithmic` environment that typesets the algorithn. But point taken about tlc3

FrankMittelbach:
ah then I remembered incorrectly and my change in tlc was wrong it should change to algorithmic then and we should alter the included-in on both.

(I don't have space on that page so can put in only one name :-) and anyway it is about making algorithms not making them float os algorithmic it is ...

FrankMittelbach:
changed the entries

PR #512, opened 2024-08-06
add bibunits compatibility

mbertucci47:
Lists [bibunits](https://www.ctan.org/pkg/bibunits) as partially-compatible because with hyperref there are no hyperlinks.

### Edit
Changed status to currently-incompatible after comment below. Also changes harvard's status to incompatible

### DIFF
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -1162,13 +1162,13 @@
 
  - name: bibunits
    type: package
-   status: unknown
+   status: partially-compatible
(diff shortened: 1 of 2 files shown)


u-fischer:
I would say not compatible. Also without hyperref the tagging is not correct as there is no Reference structure around the citation. If natbib is loaded the tagging is wrong too.

mbertucci47:
@u-fischer So then harvard (listed as partially-compatible in #469) should also be currently-incompatible because it also does not have the Reference structure, right?

Issue #443, opened 2024-08-03
ams classes' `\maketitle` errors


mbertucci47:
This may be already known but it's not mentioned in the note under amsart, etc. in the yaml file.

`\maketitle` errors in the ams classes. Example:
```tex
\DocumentMetadata
  {
    lang=en-US,
    pdfversion=2.0,
    pdfstandard=ua-2,
    testphase={phase-III,math,table,title,firstaid}
  }
\documentclass{amsart}

\title{Some title}
\author{Some author}

\begin{document}
\maketitle
text
\end{document}
```
errors with
```
! Package tagpdf Error: The number of automatic begin (3) and end (4)
(tagpdf)                text-unit para hooks differ!
```

FrankMittelbach:
I think it is generally the case that most classes fail with respect to their frontmatter handling, eg their \maketitle behavior even if they otherwise basically work. That is due to the fact that there often low-level bock manipulation happen and that causes chaos with tagging. It also probably means that many of them become compatible the moment we offer a sensible set of interfaces to set such things up.

Guess we should link that issue into the table (and perhaps add to the comment) and also check title handling in other classes, guess they are all similar. One can, kind of repair it though which is what has been done in the arxiv article that is mentioned in the yaml file if I remember correctly.

u-fischer:
With the current code in develop it not longer errors, but not everything is correct: the title is not tagged as Title, we get only a normal paragraph, and the spacing between title and author is wrong (imho as they use \trivlist which behaves now differently).   

Be aware that with the ams classes it matters if there is an empty line after the \maketitle or not.

PR #517, opened 2024-08-06
add amsproc

mbertucci47:
Adds an entry for [amsproc](https://www.ctan.org/pkg/amsproc). I kept `supported-through: [phase-III,firstaid,title]` from amsart and amsbook but it would need its own
```tex
\AddToHook{class/amsproc/after}
 {\def\@author{\authors}}
```
in latex-lab-firstaid. (I'm not sure what this does; with amsart/amsbook the metadata fields for title and author are empty. Also `\maketitle` is broken with the ams classes (#443) so listing `supported-through: [title]` might be a little misleading.)

### DIFF
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -10411,6 +10411,7 @@
    priority: 2
    supported-through: [phase-III,firstaid,title]
    comments: "theorem environments not compatible, see amsthm. Used by project example arxiv3"
+   tests: false
    tasks: needs further tests
    updated: 2024-07-07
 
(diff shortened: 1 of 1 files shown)


davidcarlisle:
leaving this one so the comment about additions to latex-lab doesn't get lost once the merged PR goes from the list

FrankMittelbach:
if the \maketitle fails I guess this os more in the area of currently incompatible, but perhaps it can be easily mended. @u-fischer any thoughts?

u-fischer:
I added a firstaid similar to the amsart one to firstaid, so they now should be similar.

u-fischer (commented):
I would extend the comment (and similar for the other ams-classes) and mention that the \maketitle command needs adjustments (both to get the look and the tagging correct). Apart from this is fine. The firstaid will be extended to cover also this class.

davidcarlisle:
I'll merge to add the amsproc entry then adjust the comments for ams classes in line with @u-fischer comment above

PR #519, opened 2024-08-06
add changebar compatibility

mbertucci47:
Lists [changebar](https://ctan.org/pkg/changebar) as partially-compatible and adds a test. The bars are not tagged as Artifacts but instead in the main stream of content.

changebar_content

This didn't adversely affect the result when uploaded to ngpdf, but may still be incorrect.

Also, it needs luatex85 with luatex. With luatex it errors with
```
! Undefined control sequence.
\cb@defpoint ...ount }}\let \the =\z@ \pdfsavepos 
                                                  \edef \cb@temp {\write \@a...
```
so it's assuming luatex defines the same pdftex primitives.

### DIFF
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -1727,7 +1727,7 @@
    priority: 2
    issues:
    tests: true
-   tasks: "Broken with lualatex."
+   tasks: "Needs luatex85 loaded with lualatex."
    updated: 2024-08-06
 
  - name: changepage
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -1722,13 +1722,13 @@
 
  - name: changebar
    type: package
-   status: unknown
+   status: partially-compatible
    included-in: [tlc3, arxiv01]
    priority: 2
    issues:
-   tests: false
-   tasks: needs tests
-   updated: 2024-07-14
+   tests: true
+   tasks: "Broken with lualatex."
+   updated: 2024-08-06
 
  - name: changepage
    type: package
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -1727,8 +1727,9 @@
    priority: 2
    issues:
    tests: true
-   tasks: "Needs luatex85 loaded with lualatex."
-   updated: 2024-08-06
+   tasks: "Determine how to tag changed parts."
+   comments: "Needs luatex85 loaded with lualatex."
+   updated: 2024-08-07
 
  - name: changepage
    type: package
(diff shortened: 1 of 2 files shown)


davidcarlisle:
I am not sure changebars should be artifacts, they are not decorative elements they may be the only indication of diff marking in a change document so it may not make sense without, leaving this one for the others to comment

car222222:
I agree with the principle behind the @davidcarlisle comment here.
But it is not the visual rules themselves that need to preserved: rather, it is important to somehow mark-up and preserve the information they are trying to point out, concerning which lines have been changed.

FrankMittelbach:
I think it is an interesting question of what should be provided. Maybe one for the PDFA LWG and UA folks to ponder?

I think we can call it partially compatible with  a `comments=` stating that how to tag changed parts to convey this information properly needs to be determined (and then implemented).

davidcarlisle:
@car222222  er yes, you wouldn't want it read as "vertical rule somewhere in the right margin" you'd want it read as "start of changed region" .... "end of changed region"  as you say, which might  means specific tags for that end points and still mark the rules as artifact.

_data/tagging-status.yml
```diff
@@ -1722,13 +1722,13 @@
...
-   status: unknown
+   status: partially-compatible
    included-in: [tlc3, arxiv01]
    priority: 2
    issues:
-   tests: false
-   tasks: needs tests
-   updated: 2024-07-14
+   tests: true
+   tasks: "Needs luatex85 loaded with lualatex."
```
FrankMittelbach, line 1730: current tasks should go into comments and instead there it should say: "determine how to tag changed parts"

FrankMittelbach (changes requested):

FrankMittelbach:
by the way, isn't the requirement that luatex85 is needed a bug of the package proper?

u-fischer:
Yes I agree that some begin/end marker are needed and that the rule should be marked as artifact. A structure begin/end will not work, as there is no guarantee that a changebar correctly nests with the rest of the structure.

car222222:
@u-fischer Ah yes, good point!
But then, the best structures are, of course, non-hierarchical -- "nesting considered harmful!"?

u-fischer:
> by the way, isn't the requirement that luatex85 is needed a bug of the package proper?

well it at least shows that the package hasn't been properly adapted to luatex. I told Johannes at TUG that one probably should modernize the package in general a bit, not only for tagging.

FrankMittelbach:
so I guess there should be an issue for this package

_data/tagging-status.yml
mbertucci47: Fixed

FrankMittelbach (approved):

PR #526, opened 2024-08-07
add hypbmsec compatibility

mbertucci47:
Lists [hypbmsec](https://ctan.org/pkg/hypbmsec) as compatible and adds a test.


FrankMittelbach:
yes, agree with you. I'll change the entry

PR #529, opened 2024-08-07
update to hypbmsec

FrankMittelbach:


_data/tagging-status.yml
```diff
@@ -4375,12 +4375,14 @@
 
  - name: hypbmsec
    type: package
-   status: compatible
+   status: partially-compatible
    included-in: [arxiv001]
    priority: 7
    issues:
+   comments: Currently compatible, but once the new tagging interface for headings is in place this might stop working
```
u-fischer, line 4382: tagging interface -> template interface?

FrankMittelbach: well it is both , but yes, definitely better

PR #535, opened 2024-08-07
add ltugboat compatibility

mbertucci47:
Lists ltugboat as compatible. It's based on the article class and seems to not do anything that messes up tagging. The commands `\bull` and `\cents` could use unicode mappings but I didn't think that was big enough to call it only partial. If it is just let me know.

### DIFF
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -10543,13 +10543,13 @@
  - name: ltugboat
    ctan-pkg: tugboat
    type: class
-   status: unknown
+   status: compatible
    included-in:
    priority: 9
    issues:
-   tests: false
-   tasks: needs tests
-   updated: 2024-07-18
+   tests: true
+   comments: "`\\bull` and `\\cents` need Unicode mappings."
+   updated: 2024-08-07
 
  - name: ltxdoc
    type: class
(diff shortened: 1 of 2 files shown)


FrankMittelbach:
agree with you. @kberry might want to know about the missing Unicode mappings but this is really a minor issue.

kberry:
I don't know how to add "Unicode taggings" to the essentially-plain-TeX
definitions of \bull and \cents (and no doubt other cmds). Please
advise.

BTW, any success of ltugboat wrt tagging is surely due to Ulrike :).

Thanks,
Karl

u-fischer:
> I don't know how to add "Unicode taggings" to the essentially-plain-TeX definitions of \bull and \cents (and no doubt other cmds). Please advise.

Well the best for `\cents` would be to use a proper glyph with `\textcent` (or `\textrm{\textcent}`). 

In cases where it is not possible to use a proper glyph with correct unicode mapping, the general idea is to surround the symbol with an ActualText, e.g. with the accsupp package:

~~~~
\usepackage{accsupp}

\makeatletter
\def\cents{\leavevmode{\BeginAccSupp{method=hex,unicode,ActualText=00A2}%
 \rm\raise.2ex\rlap{\kern.05em$\scriptstyle/$}c\EndAccSupp{}}}

\def\bull{\leavevmode\BeginAccSupp{method=hex,unicode,ActualText=25AA}%
          \vrule \@height 1ex \@width .8ex \@depth -.2ex
          \EndAccSupp{}}
~~~~

But practically it doesn't work always. E.g. despite the ActualText I can't copy & page the \bull in adobe reader - imho the problem is that there is only a rule and no glyph.

PR #560, opened 2024-08-13
add flippdf compatibility

mbertucci47:
Lists [flippdf](https://ctan.org/pkg/flippdf) as compatible and adds a test.

### DIFF
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -3442,13 +3442,12 @@
 
  - name: flippdf
    type: package
-   status: unknown
+   status: compatible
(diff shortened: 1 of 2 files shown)


u-fischer:
hm, when one upload that to ngpdf.com the order of the (mirrored) text looks rather random. But it looks more like an ngpdf bug. But on the other side I doubt that anybody wants tagging when using this package ....

FrankMittelbach:
but if there is a bug with ngpdf, we should tell Boris even if the package as such isn't really useful with tagging.

PR #567, opened 2024-08-13
add dblfloatfix compatibility

mbertucci47:
Lists [dblfloatfix](https://ctan.org/pkg/dblfloatfix) as compatible and adds a test. It overwrites some kernel commands, though. Not sure if that will mess things up in the future.

### DIFF
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -2366,7 +2366,8 @@
    priority: 6
    issues:
    tests: true
-   updated: 2024-08-13
+   tasks: "Recheck when OR hooks become available"
+   updated: 2024-08-16
 
  - name: dblfnote
    type: package
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -2361,13 +2361,12 @@
 
  - name: dblfloatfix
    type: package
-   status: unknown
+   status: compatible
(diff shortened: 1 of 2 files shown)


davidcarlisle:
Added @FrankMittelbach  to review, marking any package that overwirtes bits of the output routine as "compatible" seems inherantly risky even if the tests show no problems so far.

FrankMittelbach:
I really think we should start adding another key, maybe called `area:` (or a different name if any of you can think of any) and give it values like `OR`, `footnote`, `table`, `heading` so that we can easily identify them as something to recheck, when we do add sockets and hooks to these areas. E.g., yes, a packages that hooks into the OR needs adjustments when we provide suitable hooks and sockets there, because at that point it will either break or it will break our stuff and instead it should then use the new hooks and sockets.

Perhaps, it is too much, but I think a pass through the already collected packages and at least classify those that fall into these four categories would prove to be helpful in the end.

FrankMittelbach (approved):
I think compatible is ok, but with a task note saying "recheck when OR hooks become available"  and/or a general "area" key as suggested (but that would require going over all packages)

mbertucci47:
Task note added

davidcarlisle:
thanks, merging....

PR #580, opened 2024-08-14
corrections and additions

mbertucci47:
Corrects priority in a few places.

Adds apptools, dashbox, fp, halloweenmath, makebox, ncccropbox, ncccropmark, showlabels, svn-prov, and tabstackengine without commenting on status (all arxiv001 or above).

Lists eqnarray as incompatible with a test.

### DIFF
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -3802,6 +3802,16 @@
    tests: false
    updated: 2024-07-21
 
+ - name: fp
+   type: package
+   status: unknown
+   included-in: [arxiv01]
+   priority: 6
+   issues:
+   tests: false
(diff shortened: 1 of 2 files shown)


_data/tagging-status.yml
```diff
@@ -4258,6 +4297,16 @@
    tests: false
    updated: 2024-07-21
 
+ - name: halloweenmath
+   type: package
+   status: unknown
+   included-in: [arxiv001]
```
FrankMittelbach, line 4303: that's funny that this shows up in arxiv001 but I believe you

mbertucci47: I thought so too :)

PR #593, opened 2024-08-15
add showlabels compatibility

mbertucci47:
Lists [showlabels](https://ctan.org/pkg/showlabels) as compatible and adds a test. The labels are correctly tagged as artifacts.

### DIFF
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -8462,13 +8462,12 @@
 
  - name: showlabels
    type: package
-   status: unknown
+   status: compatible
    included-in: [arxiv001]
    priority: 7
    issues:
-   tests: false
-   tasks: needs tests
-   updated: 2024-08-14
+   tests: true
+   updated: 2024-08-15
 
  - name: showkeys
    type: package
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -8462,11 +8462,12 @@
 
  - name: showlabels
    type: package
-   status: compatible
+   status: partially-compatible
    included-in: [arxiv001]
    priority: 7
    issues:
    tests: true
+   tasks: "Determine how to tag the shown labels, matching showkeys."
    updated: 2024-08-15
 
  - name: showkeys
--- a/tagging-status/testfiles/showlabels/showlabels-01.tex
+++ b/tagging-status/testfiles/showlabels/showlabels-01.tex
@@ -1,0 +1,30 @@
+\DocumentMetadata
+  {
+    lang=en-US,
+    pdfversion=2.0,
+    pdfstandard=ua-2,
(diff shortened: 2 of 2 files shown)


FrankMittelbach:
I think it is arguable if this is correct. Artifact means that this information should be ignored when presenting the document in a different format. However, one can argue that this is intentional information, ie data dat should be presented, e.g., as an "aside". I think this warrants further thought.

FrankMittelbach:
In other wirds, maybe this should be labeled, partially compatible with a note or task stating that what kind of tagging should be deployed needs further thoughts.

_data/tagging-status.yml
```diff
@@ -8462,13 +8462,12 @@
 
  - name: showlabels
    type: package
-   status: unknown
+   status: compatible
```
davidcarlisle, line 8465: @FrankMittelbach yes someone using show lablels to edit a document would want them read out so probably not an artifact. I note `showkeys` is marked as partially-compatible with a comment querying the tagging status. In the end they should both end up with similar tagging.  perhaps make tis partially-compatible to match.

mbertucci47:
Changed to partially-compatible with a tasks entry. Let me know if anything should be changed or added

FrankMittelbach:
@mbertucci47 since the discussion we had is here and not on any issue that can be easily found, either link the merge record or not in thecomments that these need some decision on the right tagging, thanks

nxg:
I'm the showlabels maintainer, and would be very happy to make suitable adjustments to the package, if someone can give me a suitable pointer, to get me started (@davidcarlisle, _of course_ I'm not going to skip the opportunity to do so, not if `showkeys` does it properly!)

I'm going to be spending some $WORK time on learning about PDF tagging of LaTeX documents, in pursuit of accessibility goals, so I'm doubly motivated here, as a fix-up project such as this would be a very useful on-ramp to the relevant documentation and tooling.

u-fischer:
@nxg well at first the claim above that the labels are tagged as artifacts is wrong. This is only the case if lualatex is used, with pdflatex the labels are untagged and that should probably be changed (or documented that with pdflatex the document is not fully tagged).

I'm not sure if tagging these keys as Aside as suggested above is really a good idea. I don't quite see why e.g. a blind person should use a screen reader to find the labels in a document if they can use an editor and search. And if they do, they probably want that read directly at the relevant structure, e.g. inside the heading and not somewhat later. Personally I think keeping the current status would be ok unless some concrete request for a change comes up. 

But there should be more tests with different variants. Most importantly (as you hook e.g. into `\maketag@@@`) there should be tests with math tagging to ensure that this doesn't break something, so with `\DocumentMetadata{tagging=on,tagging-setup={math/setup=mathml-SE}}`  and unicode-math.

PR #603, opened 2024-08-16
add axodraw2 compatibility

mbertucci47:
Lists [axodraw2](https://ctan.org/pkg/axodraw2) as compatible and adds a test. Because the definition of `axopicture` is in terms of `picture` it can pick up the new optional argument to set alt text or artifact (of course, this is not documented anywhere).

### DIFF
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -870,6 +870,8 @@
    priority: 9
    issues:
    tests: true
+   comments: "`axopicture` environment can take same optional argument as `picture`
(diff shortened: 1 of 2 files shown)


FrankMittelbach:
> it can pick up the new optional argument to set alt text or artifact (of course, this is not documented anywhere).

you might want to add that as a remark in the comments precisely because it isn't necessarily obvious and not  documented.

mbertucci47:
Comment added

FrankMittelbach (approved):

PR #620, opened 2024-08-19
add smartref compatibility

mbertucci47:
Lists [smartref](https://www.ctan.org/pkg/smartref) as partially-compatible because it doesn't produce links with hyperref. The `\byshortname` and `\byname` commands also produce `Suppressing link with empty target` warnings with hyperref.

### DIFF
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -8573,6 +8573,16 @@
    tests: true
    updated: 2024-07-26
 
+ - name: smartref
+   type: package
+   status: partially-compatible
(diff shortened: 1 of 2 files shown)


davidcarlisle:
"no links with hyperref" is verging towards `currently-incompatible` ?

mbertucci47:
Well it never had links with hyperref to begin with so I was thinking partial since loading the tagging code is not breaking anything that already worked.

davidcarlisle:
sounds a reasonable argument:-)

PR #644, opened 2024-08-21
add poetry compatibility

mbertucci47:
Lists [poetry](https://www.ctan.org/pkg/poetry) as partially-compatible because the tagging of each line and line number as its own `<Part>` is not ideal.

### DIFF
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -7837,6 +7837,16 @@
    tests: true
    updated: 2024-07-26
 
+ - name: poetry
+   type: package
+   status: partially-compatible
+   included-in:
+   priority: 9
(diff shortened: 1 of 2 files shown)


car222222:
Why does Part get used here?

FrankMittelbach:
It is `<text-unit>` that each line is mapped to (ie each line is a single paragraph in LaTeX's eyes. That in turn in roll mapped to Part because all "TeX" paragraphs are by default something like
```
<text-unit>
   <text> ... </text>
   <text> ... </text> % or list or ....
</text-unit>
```
which roll map to
```
<Part>
  <p> .... </p>
  ...
</Part>
```
basically because `<P>` is unable to represent a real paragraph in many cases.

PR #646, opened 2024-08-21
add eepic compatibility

mbertucci47:
Lists eepic as compatible, though it only works with latex+dvips/dvipdfmx or xelatex, neither of which give good tagging. So let me know if status should be different.

### DIFF
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -2753,13 +2753,13 @@
 
  - name: eepic
    type: package
-   status: unknown
+   status: compatible
    included-in: [tlc3, arxiv01]
    priority: 2
    issues:
-   tests: false
-   tasks: needs tests
-   updated: 2024-07-14
+   tests: true
+   comments: "Does not work with pdflatex or lualatex."
+   updated: 2024-08-21
 
  - name: eforms
    type: package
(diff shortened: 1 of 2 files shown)


_data/tagging-status.yml
```diff
@@ -2753,13 +2753,13 @@
...
-   status: unknown
+   status: compatible
    included-in: [tlc3, arxiv01]
    priority: 2
    issues:
-   tests: false
-   tasks: needs tests
-   updated: 2024-07-14
+   tests: true
+   comments: "Does not work with pdflatex or lualatex."
```
FrankMittelbach, line 2761: I thought it would work with pdflatex

mbertucci47: For me the eepic-specific commands like `\ellipse` don't show in the pdf with pdflatex

davidcarlisle: yes that's true. according to the pdftex manual Hans Hagen had some macros (probably similar to the ones graphicx still uses for metapost) to directly interpret tpic specials, which would allow this to work but by now it's probably better to let this fall into obscurity rather than dig out the old pdftex support.

davidcarlisle:
I think `compatible` is OK, as you say, adding tagging doesn't break anything not already broken.

davidcarlisle:
> I know remember what I miss-remembered. It is mentioned in TLC3 only in passing that there is no another package as a replacement. So we should drop the TLC3 label and adjust the priority

OK I can do that locally as this has been merged, no need for another PR I think

PR #658, opened 2024-08-22
add svg compatibility

mbertucci47:
Lists [svg](https://ctan.org/pkg/svg) as partially-compatible because it doesn't have a user interface to set alt text, actualtext, or artifact status. It uses `\includegraphics` for the converted image so something like this works:
```tex
\includesvg[
  pretex=\setkeys{Gin}{alt=example}
  ]{example}
```


car222222:
This is another area that will need new ideas at some stage: 
are there currently any general methods or Assistive Software for making the SVG code directly accessible (whatever that might mean) without any need for the explicit addition of Alt  text?

davidcarlisle:
@car222222  this for example https://www.w3.org/TR/SVG-access/

car222222:
Yes indeed, very interesting.  But note that this is only a suggestion for a format: not an application!

Issue #698, opened 2024-09-04
`\printacronyms` errors with block environment


mbertucci47:
If a document contains a block environment like `quote` or `equation`, and `\printacronyms` of the [acro](https://www.ctan.org/pkg/acro) package is preceded by a blank line, then you get a "number of begin and end" error. The error only seems to appear with the default `template=description`.
```tex
\DocumentMetadata
  {
    lang=en-US,
    pdfversion=2.0,
    pdfstandard=ua-2,
    testphase={phase-III,math,table,title,firstaid}
  }
\documentclass{article}
\usepackage{acro}
\DeclareAcronym{cd}{
  short = CD ,
  long = compact disc
  }
\begin{document}

\begin{quote} % or equation, etc.
Some text
\end{quote}
\ac{cd} % removing the next blank line makes the error disappear

\printacronyms
%\printacronyms[template=tabular] % this is fine

\end{document}
```
There's no problem without the tagging code.

u-fischer:
Compiles fine for me with -dev (and also with the new tagging31-XXX  branch). So I think we can call that fixed-in-release.

mbertucci47:
Ah sorry. I still need to get in the habit of using -dev

PR #699, opened 2024-09-04
add acro compatibility

mbertucci47:
Lists acro as compatible and adds a couple of tests. I wasn't sure if I should link to #698 because that issue is fixed in the -dev release.

### DIFF
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -179,9 +179,9 @@
    status: compatible
    included-in: [tlc3, arxiv01]
    priority: 2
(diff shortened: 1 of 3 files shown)


FrankMittelbach:
I don't think it hurts to list resolved issues, guess there are cases where you want to review related problems even if current version has them resolved. So I would add them.

FrankMittelbach (changes requested):
suggest to add known issues even if already resolved

mbertucci47:
Issue has been referenced

Issue #713, opened 2024-09-23
luatexja classes incompatible


mbertucci47:
The [luatexja](https://www.ctan.org/pkg/luatexja) classes have several incompatibilities with the tagging code.

The classes modeled after the standard classes, ltjarticle, ltjbook, and ltjreport, break TOC tagging, and the in the latter two `\chapter` errors.
```tex
\DocumentMetadata
  {
    lang=en-US,
    pdfversion=2.0,
    pdfstandard=ua-2,
    testphase={phase-III,math,title,table,firstaid}
  }
\documentclass{ltjreport}

\begin{document}

\tableofcontents

%\chapter{Chapter} % errors
\section{Section}
\subsection{Subsection}
text

\end{document}
```
The vertical classes, ltjtarticle, ltjtbook, and ltjtreport, have the same problems as the ltj... classes.

ltjsarticle, ltjsbook, ltjskiyou, ltjspf, and ltjsreport break tagging of section headings by redefining `\@startsection`. They also don't tag the TOC correctly (but in this case no warnings).
```tex
\DocumentMetadata
  {
    lang=en-US,
    pdfversion=2.0,
    pdfstandard=ua-2,
    testphase={phase-III,math,title,table,firstaid}
  }
\documentclass{ltjsreport}

\begin{document}

\tableofcontents

\chapter{Chapter}
\section{Section}
\subsection{Subsection}
text

\end{document}
```

Issue #87, opened 2024-06-27
luatexja breaks table tagging


u-fischer:
luatexja patches and overwrite tabular/array commands and that breaks the table tagging:
~~~~
\DocumentMetadata{
  lang        = en,
  pdfversion  = 2.0,
  pdfstandard = ua-2,
  testphase   =
   {phase-III,table}
}
\documentclass[10pt,a4paper]{report}
\usepackage{luatexja}
\begin{document}
blub
  \begin{tabular}{|l|p{3cm}|}
  hello & hola
  \end{tabular}
\end{document}
~~~~

~~~~
Package tagpdf Warning: Parent-Child 'P/pdf2' --> 'TR/pdf2'.
(tagpdf)                Relation is not allowed (struct 6, /text --> struct 7)
(tagpdf)                on line 54
~~~~

![image](https://github.com/latex3/tagging-project/assets/4047173/f7d47390-cda2-4cd5-a533-c4f80edf109a)

mbertucci47:
It seems this was fixed in https://github.com/luatexja/luatexja/commit/4e0e008cf4b6a2392b01320e9547218ca82c1830

PR #718, opened 2024-09-28
update luatexja entry

mbertucci47:
Removes the issue about table tagging from the luatexja entry and mentions that the luatexja classes are incompatible as reported in #713. Tests added as well.

By the way, should #87 be closed?

### DIFF
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -6543,7 +6543,7 @@
    status: currently-incompatible
    included-in: [tlc3]
    priority: 2
-   comments: "ltj<name>, ltjt<name>, and ltjs<name> classes incompatible."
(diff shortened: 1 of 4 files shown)


_data/tagging-status.yml
```diff
@@ -6543,9 +6543,9 @@
    status: currently-incompatible
    included-in: [tlc3]
    priority: 2
-   comments: overwrites tabular/array commands
-   issues: [87]
-   updated: 2024-07-06
+   comments: "`ltj<name>`, `ltjt<name>`, and `ltjs<name>` classes incompatible."
+   issues: [87,713]
```
FrankMittelbach, line 6547: if #87 is resolved we should perhaps stop listing it

mbertucci47: Should be removed now

PR #766, opened 2024-11-19
update pstricks status

mbertucci47:
In [this update](https://tug.org/svn/texlive/trunk/Master/texmf-dist/tex/generic/pstricks/pstricks.tex?r1=71752&r2=72868&sortby=rev&sortdir=down) pstricks added the `alt`, `actualtext`, `artifact`, and `correct-BBox` keys. Like luamplib, these use the internal tagpdf commands. The `pspicture`s are also not tagged as figures. Not sure what the correct tagging is here so changed it to partially-compatible.

### DIFF
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -8902,14 +8902,15 @@
 
  - name: pstricks
    type: package
-   status: currently-incompatible
+   status: partially-compatible
    included-in: [tlc3, arxiv1]
    priority: 2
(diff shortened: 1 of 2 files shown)


u-fischer (approved):
The description for the tagging-status page looks ok to me. It should also mention that pstricks does not support the key `artifact` (`tag=artifact` will long term either disappear, or only be needed if one wants an explicit Artifact structure). We will have to discuss with the maintainer about the implementation of the keys and which additional interfaces are needed.

PR #775, opened 2024-12-21
Namespace

davidcarlisle:
This adds a RelaxNG schema for the file format used to supply MathML to LaTeX for tagging.

All elements are in either the HTML or MathML Namespace

All the `-mathml.html` files in the example collections are adjusted to use namespaced elements.

The schema is currently  extended to allow some non-core mathml constructs.

All supplied mathml is now valid except the `+`-prefixed lengths generated for `<mpadded>` by luamml

```
nested-1/nested-1-luamml-mathml.html:108:1: error: attribute ^lspace with invalid value "+4.981pt
```


car222222:
Not sure I can parse the introduction here!  

Should I try again?

zauguin:
Regarding the `+` prefixed lengths: It would be easy enough to omit the `+` there, but why is that needed? Is this just a restriction of the schema or is there a requirement in the specs somewhere that no explicit `+` is allowed?

davidcarlisle:
@zauguin They are made invalid because unfortunately they couldn't be added to mathml-core. (sorry I thought I'd raised a luamml issue but seems I haven't)

in mathml 1-3 mpadded has a special custom length syntax where `width="10pt"` means 10pt wide, but `width="+10pt` means 10pt wider than natural width.  

MathML-Core is heavily based on a CSS layout model and the costs of not using "any CSS length" was considered too high, so all length attributes take CSS length syntax with the implied CSS meaning, which means that prefix `+` is optional but ignored.

Having the same syntax mean different things in mathml full and core is clearly problematic so the schema makes them invalid.

More details at

https://github.com/w3c/mathml/issues/103

davidcarlisle:
Sigh. This branch (and so this PR) have been extended not quite intentionally  to also add schemas for the xml files extracted from a modified version of `show_pdf_tags.lua`  Those schema are separate from the schema added for the `mathml.html` files for which this branch and PR were originally generated.

We could in theory separate these in git to make two PR, but assuming we are going to merge some version anyway I would suggest these are merged as-is from this PR. Any further updates to the structure tree schema (and more are expected) could be done then as separate PR or commits to main.

@u-fischer @FrankMittelbach @zauguin any objection to merging this?

PR #779, opened 2025-01-06
remove stray sectioning commands

u-fischer:
Removes warnings about faulty sectioning nesting from the silvanusthompson example (a subsection command inside a minipage and a paragraph in the middle of a list). The empty lines before mainmatter and backmatter are only needed temporarly until the bug in the sec code is corrected.

### DIFF
--- a/project-examples/SilvanusThompson/33283-tagged.tex
+++ b/project-examples/SilvanusThompson/33283-tagged.tex
@@ -264,8 +264,10 @@
     \small
     \phantomsection
     \pdfbookmark[0]{Transcriber's Note}{TransNote\themytnote}
-    \subsection*{\centering\normalfont\scshape%
-      \normalsize\MakeLowercase{\TransNote}}%
+    %\subsection*{
+    \centering\normalfont\scshape%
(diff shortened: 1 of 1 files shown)


davidcarlisle (approved):

FrankMittelbach (commented):
not sure how this should be used later on, but if we make changes to  external fles, I think we should put in a comment line, that the faulty structure has been corrected in front of all such places, no?

davidcarlisle:
> not sure how this should be used later on, but if we make changes to external fles, I think we should put in a comment line, that the faulty structure has been corrected in front of all such places, no?

some of the files have changed more than others. I've tried to add comments for major changes, but also every tex file was checked in before any changes made, so a git diff is the most accurate check of all changes.

FrankMittelbach:
> some of the files have changed more than others. I've tried to add comments for major changes, but also every tex file was checked in before any changes made, so a git diff is the most accurate check of all changes.

for us the git history might be fine, but for somebody looking just at the .tex file it might be helpful if the source points out why and what was changed (do to errors in the source), but it was just a suggestion in passing. I'm ok with that being held in the git history too.

PR #799, opened 2025-02-12
Add test files for asmeconf

John02139:
Adds test files for asmeconf.cls


FrankMittelbach:
thanks a lot, very much appreciated. 

PS, if you happen to have a few more cycles and want to help further, similar  thorough samples for  other classes that are listed as compatible (like the std article/book/report) would be nice.

FrankMittelbach:
If validated with https://dev.verapdf-rest.duallab.com/ against UA-2 it shows 4 errors, guess they should be looked at.

Random note: The `\EntryHeading{Roman letters}` tags a little strangely because it is a list item with an empty Lbl, but probably ok. I haven't looked closely at the document content.

@u-fischer any comments from you?

FrankMittelbach:
Hi John,

if you reply by email rather than via the web interface of github please, drop all the unnecessary clutter as this otherwise ends in the issue and makes it less readable.

> Thanks for looking at this. asmeconf validates as A-4F only, not UA-2. (The other upload, mitthesis meets both standards). The main issue I’ve seen with UA-2 under pdftex relates to caption placement in subfigures (the subcaption/caption packages, I assume). Advice is very welcome. 

In that case it shouldn't be advertised as compliant (as you ctan announcement said), but only as partially compliant (with a note which constructs do not work (yet). It is correct that subcaptions aren't yet supported so if the class uses that then documents using that part would  fail UA. So we should change the table entry to partially and add a note.

> The class does not use unicode-math at present, so I have not pursued lualatex validation at this time. (Long story, but on my to do list for the future.) I had not noticed the \EntryHeading tag. It’s a subheading in a list.

The `\EntryHeading`was just a note in passing. I don't think it is totally wrong, but it is certainly just coded as a list entry not as a subheading of a list, so for somebody using some AT it will give suboptimal information.

John02139:
Hi Frank, 

I agree that the table should be updated, will do in a minute.

What I put in the CTAN announcement was the following, which I thought was clear about where it stands:

"An experimental (and undocumented) option [captionpatch] supports tagging of the 
caption package, which is loaded by the class. With this option and current 
LaTeX tagging support, asmeconf pdfs can validate as PDF 2.0/A-4F provided that 
[spanish] option is omitted. (PDF 2.0/UA-2 validation is not possible at 
present.)"

FrankMittelbach:
> What I put in the CTAN announcement was the following, which I thought was clear about where it stands:

it is clear. Either I misremembered what I saw or I confused myself reading /A-4F as /UA-2 because I was expecting it given that the mitclass was announced as compatible. So yes, all that remains is some update to the table entry for now. Sorry for the confusion.

PR #800, opened 2025-02-12
Update tagging-status.yml

John02139:
Update testing status of asmeconf and mitthesis


FrankMittelbach:
Not that it matters much but you could have put that as part of the other pull request since it is definitely part of it.

PR #805, opened 2025-02-27
tagging status of eqnlines

u-fischer:


FrankMittelbach (approved):

nbeisert:
Thanks for adding the package eqnlines to the status list. Not sure whether this is an appropriate place to report an update (new issue?):
[eqnlines v0.6.1](https://ctan.org/pkg/eqnlines) ought to work fine with the current development kernel. Here is a testfile with assorted basic math environment tests (in part based on [/project-examples/align](https://github.com/latex3/tagging-project/tree/main/project-examples/align)):
[eqnlines-01.zip](https://github.com/user-attachments/files/19513841/eqnlines-01.zip)
Few minor incompatibilities yet to be fixed are remarked in the file.

u-fischer:
@nbeisert  it looks quite good. Only a few labels looks wrong e.g. for the gather example and the "align with last numbering". The labels itself are outside and inside there are empty mtext container.

If you want to update the tagging status you can simply make a PR (which can contain the test file and comments.

nbeisert:
@u-fischer Okay, good. I thus submitted a PR #831.
I just could not reproduce said issue about misplaced labels. Could it be related to older or newer development versions of the various components? I used latest (pdf|lua)latex-dev compilers available in texlive and v0.7 of my package: all labels were in the corresponding mtext containers in the luamml-mathml html output file. I also cross-checked the pdf output in some online pdf tagging inspection tools that I could find. Therefore, I hope all is fine in this regard, but of course it can be investigated using the files in the new PR.

PR #814, opened 2025-03-10
yamllint

davidcarlisle:
testing the yamllint action


FrankMittelbach (approved):
I think the lineending or the indentation or both have changed, as a result the yaml file is 100% diff and one can't really see what has change. If I say ignore space difference then the change is zero more or less, but that doesn't mean that you have added or deleted spaces in places where they shouldn't have changed.

So I can only trust it I guess.

davidcarlisle:
yes sorry the commit 2320404 removed the first column from (almost) the entire file, other than that the only changes were 2 places that had two spaces after the colon made one space to match all th erest.

PR #823, opened 2025-04-06
add status for box packages

mbertucci47:
Lists `adjcalc` as compatible since it's just a package for calculating lengths.

Lists `adjustbox` as currently-incompatible since some keys rely on varwidth, it doesn't register the tagging-necessary graphics keys, and the tagging for boxes in general is not great at the moment.

Lists `collectbox` as partially-compatible since sometimes spaces are lost.

Lists `easyfig` and `gincltex` as currently-incompatible since they're built on top of `adjustbox`.

Lists `realboxes` as partially-compatible since mostly it look okay, but `\Parbox` produces parent-child warnings and the graphics and dashbox macros produce wrong tagging.

Lists `storebox` as currently-incompatible since it loses all spaces in the tags.

Lists `trimclip` as partially-compatible since it depends on collectbox and `\clipbox` produces some extra empty tags.

### DIFF
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -228,13 +228,12 @@
 - name: adjcalc
   ctan-pkg: adjustbox
   type: package
-  status: unknown
+  status: compatible
   included-in: [arxiv1]
   priority: 5
   issues:
   tests: false
-  tasks: needs tests
-  updated: 2024-08-01
+  updated: 2025-04-05
 
 - name: adjmulticol
   type: package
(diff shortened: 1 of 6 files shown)


_data/tagging-status.yml
```diff
@@ -11008,13 +11022,13 @@
 - name: trimclip
   ctan-pkg: adjustbox
   type: package
-  status: unknown
+  status: partially-compatible
   included-in: [arxiv1]
   priority: 5
+  comments: "Depends on partially-compatible collectbox. `\clipbox` adds empty tags."
```
davidcarlisle: the yaml validator is complaining here: we haven't been entirely consistent but I guess this should be `\\c` the backticks affect the display when this yaml is pulled in to the markdown but don't quote the backslash within the yaml.

davidcarlisle (changes requested):

_data/tagging-status.yml
davidcarlisle: oh you fixed it:-)

davidcarlisle (approved):

PR #827, opened 2025-04-07
update babel listings in tagging-status.yml

mbertucci47:
Updates the `ctan-pkg` entries for the babel listings. Also adds georgian, japanese, mongolian, piedmontese, and thai. These are all documented in the babel manual.

For test files, the entry names may have to be changed from `babel/<language>` to something else since the testfile folder name can't contain a slash (at least on my computer). Unless it's smart enough to recognize a subfolder `dutch` in the testfile folder `babel`, I didn't try.

### DIFF
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -986,7 +986,7 @@
   updated: 2024-09-01
 
 - name: babel/afrikaans
-  ctan-pkg: babel
+  ctan-pkg: babel-dutch
   type: package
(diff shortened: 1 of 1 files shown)


mbertucci47:
There are a few other languages with ldf's mentioned in the babel manual that I didn't list because I think the code is really outdated: arabic, farsi, hindi, pinyin, and spanglish (which self identifies as obsolete).

PR #839, opened 2025-05-05
update cistercian status

samcarter:


_data/tagging-status.yml
```diff
@@ -2748,6 +2748,12 @@
   tasks: needs tests
   updated: 2024-07-18
 
+- name: cistercian
+  type: package
+  status: compatible
+  supported-through: latex-lab-testphase-tikz
```
mbertucci47, line 2754: Based on other entries I think this line should be
```yaml
supported-through: [phase-III,tikz]
```

mbertucci47:
Maybe you could add an example file in https://github.com/latex3/tagging-project/tree/main/tagging-status/testfiles showing how to use the `actualtext` key with `\cistercian`?

samcarter:
> Maybe you could add an example file in https://github.com/latex3/tagging-project/tree/main/tagging-status/testfiles showing how to use the `actualtext` key with `\cistercian`?

It get automatically used if available

```
% !TeX TS-program = lualatex-dev -interaction=nonstopmode % | txs:///view-pdf | txs:///view-log

\DocumentMetadata{testphase=latest}
\documentclass{article}
\usepackage{cistercian}

\begin{document}
\cistercian{42}
\end{document}
```

will result in 

```
+<Document
  xmlns="http://iso.org/pdf2/ssn"
  id="ID.00002"
>
  +<Part
    xmlns="http://iso.org/pdf2/ssn"
    id="ID.00005"
    xmlns:orig-ns="https://www.latex-project.org/ns/dflt"
    rolemapped-from="orig-ns:text-unit"
  >
    +<P
      xmlns="http://iso.org/pdf2/ssn"
      id="ID.00006"
      xmlns:Layout="http://iso.org/pdf/ssn/Layout"
      Layout:TextAlign="Justify"
      xmlns:orig-ns="https://www.latex-project.org/ns/dflt"
      rolemapped-from="orig-ns:text"
    >
      +<Span
        xmlns="http://iso.org/pdf2/ssn"
        id="ID.00007"
        actualtext="42"
      >
        <?MarkedContent page="1" ?>
      </Span>
    </P>
  </Part>
</Document>
```

mbertucci47:
@samcarter Ah thanks. That might be useful to mention in a `comments` entry in tagging-status.yml

PR #858, opened 2025-05-25
Create l3build testing for tagging structure of example documents

zauguin:
Add l3build configuration to use latex3/pdf_structure to do automated test of the XML representation of the tagging of all test documents in tagging-status.

Currently no special extensions are used, the tests end with `.tex` to reduce differences.


davidcarlisle:
@FrankMittelbach  yes we should probably do both: checked in wrong xml trees probably end up being confusing.

zauguin:
@davidcarlisle 
> yes we should probably do both: checked in wrong xml trees probably end up being confusing.

I like having even wrong XML trees checked in since this makes it clearer how they are broken and also makes it more visible more these are affected by changes.

Maybe we could have both types checked in but separated? Something like

 * Unknown state: `.tex` tests with `.draft.xml` reference
 * Known good: `.lst` tests with `.struct.xml` reference
 * Known broken: `.bad` tests with `.broken.xml` references

FrankMittelbach:
> @davidcarlisle
> 
> > yes we should probably do both: checked in wrong xml trees probably end up being confusing.
> 
> I like having even wrong XML trees checked in since this makes it clearer how they are broken and also makes it more visible more these are affected by changes.
> 
> Maybe we could have both types checked in but separated? Something like
> 
>     * Unknown state: `.tex` tests with `.draft.xml` reference
> 
>     * Known good: `.lst` tests with `.struct.xml` reference
> 
>     * Known broken: `.bad` tests with `.broken.xml` references

how should that work? I mean conceptually?

You would manually change the name? but the test would still work?

 
> or three testfiles directories to allow running them separately?

I think it would be simpler do do it this way. It still means some manual moving, but if we then all all 3 test dirs we would even notice if a formally broken file suddenly changes (for the better hopefull).

FrankMittelbach:
Hi @zauguin, what is the status here?

PR #860, opened 2025-05-26
update tagging-status

mbertucci47:
Adds some issues for the amsmath entry and updates the amscd and amscdx entries due to https://github.com/latex3/unicode-math/issues/657.

### DIFF
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -439,14 +439,14 @@
 
 - name: amscdx
   type: package
-  status: compatible
+  status: partially-compatible
   included-in:
   priority: 9
   supported-through: [phase-III,math]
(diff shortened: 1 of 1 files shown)


_data/tagging-status.yml
```diff
@@ -476,9 +477,9 @@
   supported-through: [phase-III,math]
   comments: "Use of math tagging currently requires support from external tools."
```
FrankMittelbach, line 478: is that statement (here and elsewhere) still true this way, or would it need to get adjusted? With luamml it should work, without any external tools, or not? @davidcarlisle ?

mbertucci47: From my limited understanding I'd think it can be removed but I'll wait for others' opinions

PR #863, opened 2025-05-28
update math package status to reflect unicode-math/luamml compatibility

mbertucci47:
Following the discussion in #860, this updates the status for several math packages to reflect compatibility with unicode-math and luamml. This only covers some packages because I want to make sure this is what was in mind before finishing the rest.

Basically any package that is incompatible with unicode-math or has no luamml support (which is most, unless they use basic constructs) is listed as partially compatible with a note.

The phrase "Use of math tagging currently requires support from external tools" was removed from those entries; I'll remove the rest in another PR once this one is merged.

Where it made sense, I added `\UseName{sys_if_engine_opentype:T}{\usepackage{unicode-math}}` to test files so that #858 doesn't error on all pdftex tests.

For the changed test files, I also used `tagging=on` instead of `testphase={...}`. I'll do a mass conversion in another PR because otherwise the relevant diffs would be difficult to find.

### DIFF
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -115,14 +115,14 @@
 
 - name: abraces
   type: package
-  status: compatible
+  status: currently-incompatible
   included-in: [tlc3, arxiv001]
   priority: 2
   supported-through: [phase-III,math]
-  comments: "Use of math tagging currently requires support from external tools."
+  comments: "No luamml support."
   issues:
   tests: true
-  updated: 2024-07-29
+  updated: 2024-05-28
 
 - name: abstract
   type: package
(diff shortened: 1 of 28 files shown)


FrankMittelbach (approved):
looks fine to me. @davidcarlisle ?

Concerning `sys-if-engine_opentype:T` Didn't we plan or even have a CamelCase name for that as well?

davidcarlisle:
> @FrankMittelbach I'm not aware of any 2e name for `\sys_if_engine_opentype:TF`. iftex has `\iftutex` but this seemed easier than loading another package in every test file

yes I think it's fine here, and anyway being very regular we can always change if we do add a test in the format. I'd have probably used `\ifx\Umathchar\undefined\else ...` out of habit but technically this is probably better anyway (as it avoids loading the whole package within the scope of the ifx)

PR #866, opened 2025-05-29
finish updated math package tagging status

mbertucci47:
This completes the updates started and explained in #863.


davidcarlisle:
thanks for the update

PR #867, opened 2025-05-30
add `\title` to test files

mbertucci47:
Adds a `\title` to remaining test files that were missing it. This helps the tests pass validation.


mbertucci47:
@FrankMittelbach I think it has to do with metadata. For example the PAC checker gives an error with a title in the XMP metadata

FrankMittelbach:
that's the reason for sure, what was surprising me initially is that the \title itself is already updating the metadata and not the \maketitle but it makes sense to do it this way.

PR #872, opened 2025-06-04
add more package statuses

mbertucci47:
Adds status and test files for bracealign, colorframed, dpfloat, efbox, and spalign. Lists nath as no-support.

### DIFF
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -3549,7 +3549,7 @@
   status: currently-incompatible
   included-in:
   priority: 9
-  issues:
+  issues: [873]
   tests: true
   comments:
   updated: 2025-06-04
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -2287,6 +2287,17 @@
   tests: true
   updated: 2024-07-15
 
+- name: bracealign
+  type: package
+  status: compatible
+  included-in:
+  priority: 9
+  supported-through: [phase-III,math]
+  issues:
+  comments:
+  tests: true
+  updated: 2025-06-04
+
 - name: braket
   type: package
   status: compatible
@@ -3015,6 +3026,16 @@
   issues:
   updated: 2024-07-06
 
+- name: colorframed
+  type: package
+  status: currently-incompatible
+  included-in:
+  priority: 9
+  comments: "See entry for the `framed` package."
+  issues:
+  tests: true
+  updated: 2025-06-04
+
 - name: colorspace
   type: package
   status: unknown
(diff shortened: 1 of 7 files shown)


davidcarlisle:
> By the way, I haven't been opening new issues for package incompatibilities since (1) several packages' incompatibilities come down to the same issue, and (2) they really muck up the issue tracker. In fact I was wondering if it would be more useful to close all the package incompatibility issues that the kernel team has no control over and use the [full tagging-status](https://latex3.github.io/tagging-project/tagging-status/full) as the official tracker that can be pointed to for package authors.

that probably makes sense (@FrankMittelbach ?) if we have issues for things that we can fix, also package authors can still open issues if they want to discuss any changes they or we need to make to make a package compatible, that can still be done, but as you say, it isn't clearly helpful to pre-emptively open issues for each package.

FrankMittelbach:
> > By the way, I haven't been opening new issues for package incompatibilities since (1) several packages' incompatibilities come down to the same issue, and (2) they really muck up the issue tracker. In fact I was wondering if it would be more useful to close all the package incompatibility issues that the kernel team has no control over and use the [full tagging-status](https://latex3.github.io/tagging-project/tagging-status/full) as the official tracker that can be pointed to for package authors.
> 
> that probably makes sense (@FrankMittelbach ?) if we have issues for things that we can fix, also package authors can still open issues if they want to discuss any changes they or we need to make to make a package compatible, that can still be done, but as you say, it isn't clearly helpful to pre-emptively open issues for each package.

Not sure really. We currently have

- open issues people and that may contain ints or workarounds even if we aren't solving the problem
- the official tracker (it is a good resources but it points back to the issues (but yes it could point to close ones))
- test files (for most or all issues)
- we have classifications (which I find rather useful, eg all stuff related to math or to footnotes etc ... there I'm also quite interested in the packages that are third party and I don't get a good list if they are closed, besides there is a differnce between closed because fixed and closed because external and that means I also get the bogus ones I'm no longer interested in

Now if somebody has a problem and goes to the issue page he or she only sees open issues by default so if your search for a package (given that you have a problem with it) you wouldn't find an entry if it is closed unless you change the default search and since the package has aproblem for you why searching in resolved bugs?

Bottom line I think we should keep them open if they are real bugs regardless ofthe fact that this means 280 current issues

FrankMittelbach:
For the same reason I'm not sure it really helps to combine issues due to the same problem but in different packages. One package owner might do something the other might not and then you end up with an issue that is neither resolved no unresolved. Yes it saves a bit up front, but we should have individual test files and in the long run I think it is beneficial if issues are separate by package. Of course if there is a kernel issue showing up in several packages then there should be only one.

Speaking of tests, you may have seen that there is some activity here, we will soon have automatic testing of the full collection of the test files that (mainly) you have written, which will be a big improvement. We will then also immediately get informed if a package gets more healthy (or less).

FrankMittelbach:
> Sounds good. I will go ahead and open new issues for the recently added incompatible packages.

since you have write access to the repo you could (if you like) also add the basic classifications (the yellow ones). Otherwise I do that, I have classified most already.

FrankMittelbach (approved):

PR #891, opened 2025-06-06
Pkgstatus

mbertucci47:
Adds links for issues to several entries. Adds test files for csquotes, enumerate, float, and marginalia. Updates the hyperref status to partially-compatible for #892.


FrankMittelbach (approved):
Not really worried, but maybe better (next time) update "updated" key when adding tests, or changing comments like.

PR #914, opened 2025-06-13
Status updates

mbertucci47:
Updates booktabs to compatible and improves the test file.

Updates the comment for cleveref since that issue seems fixed.

Changes fnpct to compatible because the issue it had with hyperref seems fixed.

Changes hypbmsec to compatible since it seems fine after the latex-lab-sec update the entry mentions.

Lists menukeys and letter as partially-compatible and minibox as currently-incompatible, see linked issues.

Several issues are linked to their entries.

### DIFF
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -12494,13 +12494,13 @@
 
 - name: letter
   type: class
-  status: unknown
+  status: partially-compatible
   included-in:
   priority: 2
-  issues:
-  tests: false
-  tasks: needs tests
-  updated: 2024-07-06
+  issues: [912]
+  tests: true
+  comments: "Tagging is not ideal."
+  updated: 2025-06-12
 
 - name: llncs
   type: class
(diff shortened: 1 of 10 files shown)


mbertucci47:
@FrankMittelbach Should I go ahead and merge PR's that don't contain anything questionable? I can also specifically request reviews but I have a feeling you gave me write access so some of the more tedious things don't need to be looked at :-)

FrankMittelbach (approved):

FrankMittelbach:
> @FrankMittelbach Should I go ahead and merge PR's that don't contain anything questionable? I can also specifically request reviews but I have a feeling you gave me write access so some of the more tedious things don't need to be looked at :-)

we tend to put on a reviewer between us on more or less all pulls. There are a few where I sometimes just do the fix, eg a single typo and similar, but on the whole  it has often helped to have a second set of eyes look at stuff even if only briefly. But we don't have  a hard rule, so use your judgement :-)

To take this PR as an example, most of the changes are obvious but there are so many and also new classifications that I if I would have done it, I guess would have bothered David or Ulrike (or you) to take a second look

Issue #875, opened 2025-06-04
`smallmatrix` luamml issue


mbertucci47:
The mathml for amsmath's `smallmatrix` seems fine to me but the log contains
```
==============...e/2025/texmf-dist/tex/lualatex/luamml/luamml-amsmath.lua:83: no
 string to print
stack traceback:
	[C]: in function 'texio.write_nl'
	...e/2025/texmf-dist/tex/lualatex/luamml/luamml-amsmath.lua:83: in upvalue 'deb
ug_mtable'
	...e/2025/texmf-dist/tex/lualatex/luamml/luamml-amsmath.lua:126: in function <.
..e/2025/texmf-dist/tex/lualatex/luamml/luamml-amsmath.lua:117>.
\__socket_tagsupport/math/luamml/mtable/smallmatrix/save_plug_luamml:w ...
                                                  
l.27 \]
       
The lua interpreter ran into a problem, so the
remainder of this lua chunk will be ignored.
```
for the example
```tex
\DocumentMetadata
  {
    lang=en-US,
    pdfversion=2.0,
    pdfstandard=ua-2,
    tagging=on,
  }
\documentclass{article}

\usepackage{unicode-math}

\begin{document}

\[
\begin{matrix}
a & b \\
c & d
\end{matrix}
\]

\[
\begin{smallmatrix}
a & b \\
c & d
\end{smallmatrix}
\]

\end{document}
```
By the way, why are the amsmath patches in two locations? They are in both `luamml-patches-amsmath.sty` and `latex-lab-amsmath.ltx`. I see https://github.com/latex3/luamml/pull/7 but both files still exist and are still loaded.

u-fischer:
Oh and the debugmtable problem is already resolved in luamml, it compiles with the luamml in the UF-docu branch.

PR #918, opened 2025-06-19
add testmath as amsmath test file

mbertucci47:
Uses the file testmath.tex as a test file for amsmath. So it didn't error, I made the adjustments
- convert all `P_\X` to `P_{\X}` (for unicode-math)
- comment out redefinition of `\@makecol`
- until new luamml is released, comment out `smallmatrix` envs (#875)

I tried running it with `tagging-setup={math/setup=mathml-SE}` as well but this results in _a lot_ of warnings. When I tried to isolate the warnings in a MWE, they went away so I gave up. It is also somewhat difficult to debug because either the line numbers in the warning seem wrong or there are no line numbers in the warning.


mbertucci47:
Okay I added back the smallmatrix envs

zauguin:
We just merged the change to do automated testing, so it would be awesome if this MR could be rebased and adjusted:

 * The testfile should be moved from `tagging-status/testfiles` to `tagging-status/testfiles-broken` to indicate that the result has known issues. (or
 * A corresponding reference file should be created with `l3build -c config-broken save testmath-tagged`

mbertucci47:
@zauguin Which directory do I run the l3build command in? Also, is that the right syntax? When I do that it says the usage is wrong

FrankMittelbach:
> @zauguin Which directory do I run the l3build command in? Also, is that the right syntax? When I do that it says the usage is wrong

l3build save -c ... I guess. And in the dir where the config file is.

mbertucci47:
I'm sorry I'm still confused. When I run `l3build save -c config-broken testmath-tagged` in the directory with the file it says
```
Error: Cannot find configuration build.lua
```

zauguin:
@mbertucci47 Thanks. Could you also move the `.struct.xml` files into the amsmath directory with the test?

mbertucci47:
@zauguin Is this okay to merge?

PR #923, opened 2025-06-21
Add instructions for moving tests into categories

zauguin:


tagging-status/README.md
```diff
@@ -18,3 +18,13 @@ XML representation of structure trees in `.struct.xml` files.
...
+To move a test into another category, follow the following three step process:
+
+1. Look at the `.struct.xml` file and compare it with the `.tex` file's content. If the file looks correct and represents the content of the `.tex` document`, then the test can be moved into the `testfiles` directory. Otherwise it should go into `testfiles-broken`.
+ If you are unsure, keep the test in `testfiles-unknown` for now.
+2. Move both the testfiles ending with `.tex` and the reference file ending with `.struct.xml` into the chosen new testfiles directory. You might have to create a new directory for the package.
+3. If you are removing the last test for a package from `testfiles-unknown` (so if the directory of the package is empty afterwards), remove the directory and update the package entry in ../_data/tagging-status.yml:
+  See if any tests for the package are classified as `testfiles-broken` and if any are classified as `testfiles`. If both are present, change the line `tests: unknown` to `tests: partially-broken`. If only `testfiles-broken` contains tests for the package, change the line to `tests: broken`. If only `testfiles` contains tests, set it to `tests: true`.
```
FrankMittelbach, line 30: true / broken / partially-broken looks a bit odd  suggest to change true -> okay. Also makes more sense without explanation if you see `tests: okay` compared to `tests: true`which I misread as "there are some tests" initially.

FrankMittelbach (changes requested):
maybe not request but suggest changes

FrankMittelbach (approved):

PR #927, opened 2025-06-27
Update references with new luamml

zauguin:
With recent tagpdf and luamml updates the generated tagging structures changed.
This imports the new state and should provide a basis to check if differences are as expected.


FrankMittelbach:
Is this matching your expectations after updates to tagpdf etc @u-fischer ? If so I think we should merge so that new changes will show up not intermixed with all this.

zauguin:
> @FrankMittelbach there one or two changes in the run status where I do not know where they are from, but as this is all in testfiles-unknown, I think it is ok.

The delarray run status change is caused by the luamml update. Previously delarray crashed luatex with luamml, now the error is detected and it cleanly terminates with a regular error.

PR #934, opened 2025-06-30
Update reference files

zauguin:


FrankMittelbach:
are the remaining differences understood?

FrankMittelbach:
in other words it could be merged and we wait for future changes, correct?

PR #938, opened 2025-07-07
Update README.md

leomayer:
The entry page has been updated and restructured with a focus on enhancing the first-time user experience. I've reordered the content based on what I, as an initial visitor, would want to see first.

One specific point of concern is the placement of the "Remark" section; I'm uncertain if it should retain its prominent position at the beginning.

I've also moved the previous opening section to the bottom. My reasoning is that a new user, like myself, isn't primarily interested in implementation specifics right away; those details become more valuable after some initial interaction.I acknowledge that this reordering is likely influenced by my own user bias, as I view myself purely as an end consumer without in-depth technical knowledge, having gained my understanding primarily from various issue discussions and requests. Aside from these structural changes, there are also some minor wording adjustments.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,36 +1,51 @@
 # The LaTeX Tagged PDF repository
+## Basic example for accessible PDFs
+This basic example compiles a tagged PDF 2.0 without any errors. To fully comply with the PDF/UA-2 standard, ensure you explicitly set the document's title and language.
+
+    \DocumentMetadata{
+		lang          = en-us,
+		pdfstandard   = ua-2,
+		pdfversion    = 2.0,
+		tagging       = on,
+		tagging-setup = {math/setup=mathml-SE} 
+    }
+    \documentclass{article}
+    \usepackage{hyperref}
+    \hypersetup{
+    	pdftitle=Accessible Document Example
+    }
+    \begin{document}
+      This is a tagged PDF with some accessible content.
+    \end{document}
+
+For more detailed examples and various topics, please see the sections below.
+## Using LaTeX for accessible PDF
+Check out our documentation on [using the prototype to generate accessible PDFs](https://latex3.github.io/tagging-project/documentation/prototype-usage-instructions). The documentation will be updated whenever we alter the still temporary interfaces.
+## Documentation
+At the moment documentation about tagging is sprinkled across different places. Here is [an overview on where to find these places](https://latex3.github.io/tagging-project/documentation/documentation-pointers).
+
+Help is welcomed to combine them into a single document.
+
+Beyond the documentation, you can also find answers to various issues in " [Discussions](https://github.com/latex3/tagging-project/discussions)" or [Tex.stackexchange.com ](https://tex.stackexchange.com/questions/tagged/tagged-pdf)(searching with `[tagged-pdf])`
+## Status of LaTeX classes
+Here's a list of core and contributed LaTeX packages, showing [their compatibility with the tagging code](https://latex3.github.io/tagging-project/tagging-status/).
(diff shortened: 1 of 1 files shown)


davidcarlisle:
I agree that the README could do with some updates, but I'm not sure that this is going in quite the right direction.

I don't think it should launch straight in to a code block before any kind of introduction about what is tagged pdf or the latex tagging project, similarly mentioning technical terms like PDF/UA-2 on _line 1_ with no explanation or even a link assumes people are far more familiar with tagging and pdf specification terminology than is reasonable for the top level page.

Also I don't think the examples in discussions/72 should be relegated to a history section, the source code of those examples still forms a major part of this repository and is maintained and updated as the markup to enable tagging has been simplified at each phase of the project.

The phrasing 

> This [tagging-project repository](https://github.com/latex3/tagging-project/) does not contain any code.

> It is meant to track issues and suggestions related to the "LaTeX Tagged PDF" project.

comes from the current README and was true when written (and the repo only had issues and discussions) 
but now less so, depending how you define "code2 notably it has the source code of all the example documents for
discussions/72 and it has (by now quite a large) set of test files for the tagging-status pages.

davidcarlisle (changes requested):
See comments posted to the PR

davidcarlisle:
@leomayer  I think the README should reflect the reason for the repo which was/is as it says

> it is meant to track issues and suggestions related to the "LaTeX Tagged PDF" project. 

That means that its primary aim is to host discussions, the four most important currently are the four pinned discussions at
https://github.com/latex3/tagging-project/discussions and issues at https://github.com/latex3/tagging-project/issues

Hosting the tagging status tests is another important feature, and finally

hosting some general user documentation on tagging.


The current README doesn't really highlight these separate parts of this repository as it mostly just reflects the original use to host discussions, but the suggested rewrite is not really explaining the repository in these terms either.

leomayer:
@u-fischer Thanks for the hint with the `hyperref` and that there is an easier way to set the attributes. The mentioning of

```
tagging-setup = {math/setup=mathml-SE}
```
is more or less the trade off to have a code which compiles without warnings.

@davidcarlisle: about "Discussion" -  I've mentioned it in the `Documentation` section as last point that discussion  could take place here as well. I have honestly no clue how to make it more obvious.

One point additionally: The [page]( https://latex3.github.io/tagging-project/)  has no direct link to the github. While it might be clear for you it might not really evident for users who land on this page.

For the rest I'm quite unclear what the "main" intention of the landing page should be. Because I've read @FrankMittelbach [comment ](https://github.com/latex3/tagging-project/discussions/936#discussioncomment-13643303) little bit different then yours, @davidcarlisle. But perhaps I'm wrong.

To my understanding the "prototype" page gives already many (good) example in different areas. Therefore I would look there first for "documentation". I agree that perhaps Discussion/72 has a lot of examples and history included. I personally would hardly look into it since too technical. 

I understand that there are issues and discussions pinned - which are important. As an advanced user I would look into it and find them easily.

Which raises again the question: Who should be targeting with the landing page? Is it for "newcomers", TeX-experts, developers, those who look for documentation?

Anyway one of the reasons why I kept the old text is that I can hardly evaluate what is relevant and what not. I can only evaluate what would be important for me (as a non-tex-dev).
****

I've update the file with tiny changes.

davidcarlisle:
> Most of the texts I've copied therefore "produced by the LaTeX Project by March 2024" is just copied. 

yes sure, the Readme needed an update.

Thanks for the PR, I suspect that we'll not merge this as is but will take your points into consideration during a rewrite

PR #940, opened 2025-07-07
README - proposal

leomayer:
restored version of pull request #938 - eventual further discussions on #938


davidcarlisle:
Thanks for the suggestions here they were taken into account but the readme has been re-organised in recent PR so this will no longer merge as written so closing here.

PR #941, opened 2025-07-07
Add entries for extsizes and related classes

IanThompson3:


davidcarlisle (changes requested):
please add the missing `-` before `name: extsizes`, otherwise it looks OK, thanks

IanThompson3:
I think that's done @davidcarlisle . I'm not very good with this fancy github stuff!

davidcarlisle:
> I think that's done @davidcarlisle . I'm not very good with this fancy github stuff!

yaml is pretty weird but you get used to it:-

(by the way not all Fortran programs go undocumented https://support.nag.com/numeric/nl/nagdoc_latest/   :-)

davidcarlisle (approved):

_data/tagging-status.yml
```diff
@@ -4314,6 +4314,14 @@
   tests: unknown
   updated: 2024-08-06
 
+- name: extsizes
+  type: package
+  status: compatible
+  issues:
+  tests:
```
mbertucci47, line 4321: needs tests: false

_data/tagging-status.yml
```diff
@@ -12489,6 +12497,51 @@
   tasks: needs tests
   updated: 2024-07-17
 
+- name: extarticle
+  type: class
+  status: unknown
+  issues:
+  tests: false
+  tasks: needs tests
+  comments: "Perhaps safer to use article + extsizes in the interim."
```
mbertucci47, line 12506: Is this comment necessary? Anything from article.cls that is incompatible with tagging is redefined after the class is loaded, so I think the same should be true for extarticle

_data/tagging-status.yml
```diff
@@ -12489,6 +12497,51 @@
...
+  comments: "Perhaps safer to use book + extsizes in the interim."
+  updated: 2025-07-07
+
+- name: extletter
+  type: class
+  status: unknown
+  issues:
+  tests: false
+  tasks: needs tests
+  comments: "Perhaps safer to use letter + extsizes in the interim."
```
mbertucci47, line 12524: letter is currently incompatible

_data/tagging-status.yml
mbertucci47: Oh I see https://github.com/latex3/tagging-project/discussions/939 now. The comments should say basically what Ulrike said in the discussion so that looking back one can see what needs fixing

_data/tagging-status.yml
davidcarlisle: partially-incompatible according to status, I think leaving extletter as unknown and needs tests is OK for now. I may adjust the comment a bit

PR #949, opened 2025-07-13
custom l3build save command

davidcarlisle:
This PR installs a modified `save` command that for test files named 

_directory_ `-` _digit_ ...

copies the newly generated reference file into the subdirectory _directory_ not the top level test directory.

Test files not of that form are left in the main test directory, as now.

There are only a few tests here (amsmath) that do not fit this pattern and we could rename those to match.

This means that after a test failure if you execute the `l3build save` command shown in the workflow log, the new reference files are copied in place and `git diff` will show the differences.

### DIFF
--- a/tagging-status/build.lua
+++ b/tagging-status/build.lua
@@ -63,3 +63,59 @@ test_types = {
     end,
   },
 }
+
+-- copy saved reference file back to subdir if suitably named.
+function tagging_save(names)
+  do
+    local errorlevel = checkinit()
+    if errorlevel ~= 0 then
+      return errorlevel
+    end
+  end
+  local engines
+  if options["engine"] then
+    engines = checkengines -- sanitized by check_engines()
+  else
+    engines = {stdengine}
+  end
+  if names == nil then
+    print("Arguments are required for the save command")
+    return 1
+  end
+  for _,name in pairs(names) do
+    local test_filename, kind = testexists(name)
+    if not test_filename then
+      print('Test "' .. name .. '" not found')
+      return 1
+    end
+    local test_type = test_types[kind]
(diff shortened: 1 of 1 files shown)


zauguin (approved):

FrankMittelbach:
> ``
> 
> we could try to make the save command smarter or not use the two level structure and have
> 
> ```
> testfiles-unknown/babel-esperanto/babel-esperanto-01-test.struct.xml
> ```

I would choose the the simple solution and require naming convention, but up to you :-)

davidcarlisle:
> > ``
> > we could try to make the save command smarter or not use the two level structure and have
> > ```
> > testfiles-unknown/babel-esperanto/babel-esperanto-01-test.struct.xml
> > ```
> 
> I would choose the the simple solution and require naming convention, but up to you :-)

done, moved to `babel-esperanto` as suggested.
 `library` are not shown in the filtered list just the full list, I added a link to the full list in the babel entry so the full babel list is easily reachable.

FrankMittelbach:
well we could have them as packages & libraries in the main (filtered) table or we could have a third filtered tabled just for libraries ... or we can just leave it like that with the note (but then a similar note should go into tikz and tcolorbox the moment we do add some of their libraries).

davidcarlisle:
> well we could have them as packages & libraries in the main (filtered) table or we could have a third filtered tabled just for libraries ... or we can just leave it like that with the note (but then a similar note should go into tikz and tcolorbox the moment we do add some of their libraries).

yes that was my thought, just having the note/link for now. If we end up wanting to show "libraries" in the filtered list (or treating them as packages there) we could of course adjust the listing templating, but I think this is fine to allow the yaml data to be built up.

PR #951, opened 2025-07-13
Disable GitHub Pages site and replace with site under latex-project.org

zauguin:
Currently the page created by this repository is reachable under https://latex3.github.io/tagging-project/ and under https://tagging-project.latex-project.org/. This leads to issues since the links assume that the base URL is https://latex3.github.io/tagging-project/, so some links are leading to unexpected locations when viewed under https://tagging-project.latex-project.org/.

My suggestion is to change the configuration to fully use the https://tagging-project.latex-project.org/ URL and stop deploying to https://latex3.github.io/tagging-project/. Alternatively we could also do the opposite and remove the https://tagging-project.latex-project.org/ version again.

### DIFF
--- a/.github/workflows/pages.yaml
+++ b/.github/workflows/pages.yaml
@@ -8,7 +8,6 @@ on:
 
 permissions:
   contents: read
-  pages: write
   id-token: write
 
 # Allow only one concurrent deployment, skipping runs queued between the run in-progress and latest queued.
@@ -49,5 +48,3 @@ jobs:
         uses: zauguin/custom-pages@trunk
         with:
           server: https://deploy.latex-project.org
-      - name: Deploy to GitHub Pages
-        uses: actions/deploy-pages@v4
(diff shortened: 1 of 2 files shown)


gucci-on-fleek:
@zauguin 

> > I think we'd have to keep this with redirects
> 
> Sadly redirects in GitHub Pages don't really work (at least not HTTP redirects,

I just did this a few months ago; it's undocumented and a little annoying to set up, but it works fairly well. You need to first set the new site up as a custom domain in `https://github.com/<user>/<repo>/settings/pages`. GitHub will make you set up a CNAME record on the target domain pointing to some special GitHub domain, and afterwards it will redirect the `https://<site>.github.io` links to the new domain that you set up. But, the trick is that after you've set this up and GitHub has validated the DNS records, you can change the DNS back to whatever you want, and GitHub will continue redirecting `https://<site>.github.io` to the target domain, even if it's no longer hosted by GitHub.

Issue #357, opened 2024-07-29
asciilist package incompatibility


mbertucci47:
The [asciilist](https://www.ctan.org/pkg/asciilist) package is mostly compatible with the tagging code, however the commands `\AsciiListFromFile(s)` and `\AsciiDocListFromFile(s)` error if the document contains another list.
```tex
\DocumentMetadata
  {
    lang=en-US,
    pdfversion=2.0,
    pdfstandard=ua-2,
    testphase={phase-III,math,table,title,firstaid}
  }
\begin{filecontents*}[overwrite]{AsciiContents.example}
 * item
   - sub-item
 * another item
\end{filecontents*}

\documentclass{article}
\usepackage{asciilist}

\title{asciilist tagging test}

\begin{document}

\begin{enumerate}
\item bla
\item blub
\end{enumerate}

\AsciiListFromFile{auto}{AsciiList.example}

\end{document}
```
errors with
```
! Package tagpdf Error: The number of automatic begin (7) and end (6)
(tagpdf)                text-unit para hooks differ!
```
The definition of `\AsciiListFromFile` is just
```tex
\newcommand\AsciiListFromFile[3][]{%
  \AsciiList[#1]{#2}%
  \input{#3}%
  \endAsciiList}
```
so I think all that needs fixing is changing it to
```tex
\newcommand\AsciiListFromFile[3][]{%
  \begin{AsciiList}[#1]{#2}%
  \input{#3}%
  \end{AsciiList}}
```

mbertucci47:
I don't know what changed to remove the error but I'm going to go ahead and close this since #953 was merged.

PR #953, opened 2025-07-16
move some test files

mbertucci47:
Hopefully all obvious except asciilist where the error seems to have magically disappeared (#357). It will take a while to recheck all these tests...

I couldn't figure out why some tests have only one xml file. For example, arydshln-01 does not have a pdftex.struct.xml. Same with authblk-02 and autonum-01.


mbertucci47:
What is the correct way to update tests that are deemed okay? The asciilist tests' xml files need to be updated but the test suite is not complaining.

zauguin:
> I couldn't figure out why some tests have only one xml file. For example, arydshln-01 does not have a pdftex.struct.xml. Same with authblk-02 and autonum-01.

tl;dr: No `.pdftex.struct.xml` means that the `.pdftex.struct.xml` file would be identical to the `.struct.xml` file.

l3build usually assumes that all engines generate the same test output. So the usual case is that only one XML file exists which is generated by all engines (like for `arydshln-01`). The `.pdftex.struct.xml` is considered the special case where the engines are expected to differ, so there is an explicit override for `pdftex`.

For tagging related tests this is a bit weird since almost all tests differ, but the special case is actually the usual case.

> I am also still confused about l3build :) I ran `l3build check --show-saves` after moving all the files, but it just prints

Each of the testfile directories uses different `l3build` configurations. You can explicitly test one of the configurations with e.g. `l3build check -c config-broken`. If you just run `l3build check` without an explicit configuration you should automatically run all the tests in all the configurations, but there's a typo in `build.lua` so only the default configuration with tests in `testfiles` run. I fixed the typo in your branch, so no you should get all the tests.

> What is the correct way to update tests that are deemed okay? The asciilist tests' xml files need to be updated but the test suite is not complaining.

The problem in the `asciilist` test is that `l3build` isolates the different test runs, so the `.example` file next to the tests are not seen during test execution. The `AsciiContents.example` gets recreated through `filecontents` in the test, but the `AsciiDocList.example` is missing. This fails the compilation, so no PDF output is generated leading to the warning and the unexpected failure in the XML file.
I pushed a change into your branch to add `AsciiDocList.example` also with `filecontents` which should resolve this issue and let the test fail as expected. Now it should be possible to regenerate the XML files as usual.

zauguin (approved):

PR #1001, opened 2025-10-13
Create CONTRIBUTING.md as guideline to authors

jasperhabicht:
The idea is to have some guidelines for class or package authors about how to contribute to this repository in order to have the process as unified and smooth as possible.


FrankMittelbach:
> It might be useful to combine this with the content of https://github.com/latex3/tagging-project/blob/main/tagging-status/README.md in one location.

not sure. might be better to link from there to contributing instead, to avoid saying the same in several places (because eventually what is said then will not agree with the other place.

jasperhabicht:
> > It might be useful to combine this with the content of https://github.com/latex3/tagging-project/blob/main/tagging-status/README.md in one location.
> 
> not sure. might be better to link from there to contributing instead, to avoid saying the same in several places (because eventually what is said then will not agree with the other place.

I am unsure as well. I think, the README was originally written for another purpose namely to organize the handling by the team. Some of the contents is of course relevant also for authors, but a how-to for authors should focus on step-wise instructions, I think, in order to make it clear for them and easy for the team to handle contributions.

tagging-status/CONTRIBUTING.md
```diff
@@ -29,21 +29,27 @@ Compile the test file and check the output using software such as PAC or VeraPDF
 
 Run `l3build save <class or package name>-01` and `l3build save -e pdftex <class or package name>-01` (for multiple test files, just concatenate the names) in your working directory to create an XML file named `<class or package name>-01.struct.xml` in the `testfiles/<class or package name>/` subdirectory. Check the contents of this file as well. This file shows a combination of tagging, structure and alternative representations which can also help you to check the tagging support of your class or package. This file will also serve as test file against which the tagging status of your class or package is later checked against.  
 
+If you spot XML comments such as `<!-- UTF16 to UTF8 conversion failure -->` in the generated XML file, it is likely that `cm-super` is not installed. These comments appear as there are not Unicode mappings for bitmap fonts which are maybe used because `cm-super` is not installed.
```
FrankMittelbach: which might get used

jasperhabicht: Thanks, added this and also introduced line breaks between semantic breaks. 

Sorry, as the structure of the repository has changed, I decided to rewrite the relevant passage in the file.

jasperhabicht:
> That was how I was using the testfiles-triage, I think actually starting in
> 
>  `testfiles-unknown/yourpackage` 
> 
> should be fine

Then, this practice could/should probably be reflected in the guidelines.

But it is maybe not that important as the first steps should be made locally anyways.

PR #1002, opened 2025-10-14
change test: true to tests: unknown

mbertucci47:
`tests: true` is not a valid string and leads to broken links in the tagging-status webpage.


davidcarlisle:
thanks

PR #1003, opened 2025-10-16
Update academicons-01.pdftex.struct.xml

mbertucci47:
The package added a type1 font so no longer errors with pdftex.

Two questions that I can ask elsewhere if preferred:
1. Running `l3build save -c config-unknown -e pdftex academicons-01` took an incredibly long time. Not to actually run tex and make the xml file, but it stalled for at least 3 minutes before actually starting the process. Is this expected?
2. Many of the tests produce
```
...tex/latex/show-pdf-tags/show-pdf-tags_process_stream.lua:80: bad argument #1 to 'match' (string expected, got nil)
Error parsing marked content stream, returning empty marked content
```
What exactly does this mean? I'm guessing that show-pdf-tags can't interpret the weird type1 font content but I'm not sure.


zauguin:
> What exactly does this mean? I'm guessing that show-pdf-tags can't interpret the weird type1 font content but I'm not sure.

I fixed the error in show-pdf-tags, so the next version will no longer fail with an internal error here. But the underlying issue is that there is no ToUnicode mapping for this font. Fixing that is tricky since there is no Unicode codepoint for most of the glyphs in this font.

mbertucci47:
Thanks. I updated the tagging-status entry to reflect that.

FrankMittelbach:
> Hmm two new failures popped up that weren't there before. Different output from texosquery

which isn't really surprising is it? more surprising that it ever worked. If I run the test, I get my OS details in whic , of course wouldn't fit anybody elses details and not those on the GitHub container either. Not sure if it is possible to construct any meaningful test for this.

davidcarlisle (approved):
we need to see if we can adjust l3build to prevent copying all the tests when just testing one, but that doesn't block this PR

PR #1010, opened 2025-10-20
move some tests

mbertucci47:
Moves the amsmath test from testfiles-broken to testfiles-partial and removes the testfiles-broken directory. Also moves some tests that were in testfiles-excluded to testfiles-triage. I'm not sure why they were in testfiles-excluded to begin with. Lastly changes texosquery to `tests: false` since that's what's been decided for excluded tests.


mbertucci47:
I see the new names in the main build.lua but when github ran the tests, it didn't test any of the new directories.

PR #1012, opened 2025-10-21
remove config-broken from build.lua

mbertucci47:
Deletes `config-broken.lua` and removes it from `checkconfigs` in the main `build.lua`. For some reason the github CI is still not running all the tests (and still running the config-broken tests) but that's out of my wheelhouse.


mbertucci47:
Ah I see, https://github.com/latex3/tagging-project/blob/main/.github/workflows/tests.yaml needs to be updated. I will let someone more knowledgeable do that.

PR #1015, opened 2025-10-21
move some triaged tests; exclude packages not in texlive

mbertucci47:
Hopefully all self-explanatory. The main thing is excluding tests for packages not in texlive. The spreadtab test had to be updated since the package removed some commands and changed the syntax.


mbertucci47:
Also, it's kind of crazy that the github CI runs the full test suite on every commit. Is there any way to make it run the tests only on changed files? And then run all tests only every so often to detect unexpected changes. I know `ci-skip` is an option but for the things I am changing, I _do_ want those tests run.

FrankMittelbach:
> Also, it's kind of crazy that the github CI runs the full test suite on every commit. Is there any way to make it run the tests only on changed files? And then run all tests only every so often to detect unexpected changes. I know `ci-skip` is an option but for the things I am changing, I _do_ want those tests run.

it was thought to be a good idea at the time :-) but perhaps not, it gets a little heavy. and right now github is either not happy in general or just not happy with us ... all the runs take ages. Guess we should make a separate issue to decide what the besat approach is going forward (can you make one? ... guess I want to turn pumpkin

davidcarlisle:
shapepar was duplicated in testfiles-incompatible and testfiles-exclude, (possibly a bad merge?) I deleted the copy in incompatible. tests are still running but it's good so far

mbertucci47:
@davidcarlisle Yes sorry about that. As for the running tests, they seem to be stuck on footnpag and warpcol, which are both tests I moved in this PR. Do you know what's going on there?

PR #1019, opened 2025-10-22
add luatex status

mbertucci47:
Adds `luatex-only: true` for several tagging-status entries. Moves the luamplib and luatexja tests to the correct folders.

One thing the current setup doesn't take into account is luatex-only tests in folders different than `testfiles-compatible-luatex`. For example, the luatexja-01 tests should be okay but if the pdftex xml file is deleted then the tests will fail as long as the tests are in `testfiles-incompatible` (perhaps luatexja should actually be partial, but that's beside the point).


davidcarlisle:
> if we do this, the key should be documented on top I think

It's documented in the yaml

```
#  luatex-only        optional   boolean         true if package only works with lualatex
```

I'm not sure it really needs to be documented in the html, it's more of an internal flag to adjust the links to tests

@mbertucci47 

> One thing the current setup doesn't take into account is luatex-only tests in folders different than testfiles-compatible-luatexv


also partial, the test is

```
{%  if p.luatex-only -%}
{%-   if p.tests == "excluded" -%}
{%-     assign dir = "-excluded" -%}
{%-   elsif p.status == "compatible" -%}
{%-     assign dir = "-compatible-luatex" -%}
{%-   elsif p.status == "partially-compatible" -%}
{%-     assign dir = "-partial-luatex" -%}
{%-   elsif  p.status == "currently-incompatible" -%}
{%-     assign dir = "-incompatible" -%}
{%-   else -%}
{%-     assign dir = "-unknown" -%}
{%-   endif -%}
```

When it was just for tests I thought it wasn't worth separating the incompatible or unknown as pdftex fails weren't an issue but now it's classifying the package it is probably clearer to split them in all cases (which would also help to  split up the long incompatible and unknown llsts in the github actions matrix).

if you make a `testfiles-incompatible-luatex` you need to add a one line `config-compatible-luatex.lua`  and I'll adjust the rest to find it....

FrankMittelbach:
> It's documented in the yaml
> 
> ```
> #  luatex-only        optional   boolean         true if package only works with lualatex
> ```

I was only thinking about the yaml, but I didn't check there only looked at the diff (and thought it was new and not already previously  added

Issue #1020, opened 2025-10-23
Running biber/bibtex/makeindex for tests


davidcarlisle:
A comment by @mbertucci47 from #1006 

> One other question: what about tests that need runs of something other than latex, e.g. bibtex or makeindex? I am thinking of the many bibliography packages.

The simplest existing case is probably 

https://github.com/latex3/tagging-project/blob/main/tagging-status/testfiles-compatible/citeall/citeall-01.tex

that's a biber test file but the reference xml just shows the `[citekey]` form used when biber hasn't been called which is not necessarily wrong but probably not that useful a test of the package functionality and impact on tagging.

`l3build` has functions that can check the log and call biber if needed I just need to hook them up (or get @josephwright to remind me how best to do that:-)

mbertucci47:
@davidcarlisle Sounds good. I will look through and adjust the test names.

davidcarlisle:
yes I feared there would be a bibunits comment coming:-)

I was hoping to avoid opening every test file to look inside for configuration comments, which is why I went for just using the filename (which was already passed to the `runtests_tasks` hook.

`chapterbib` (at least) doesn't show the required commands in the log so I think the test will need to supply them.

Perhaps instead of using separate `-bibtex` and `-biber` (and potentially `-makeindex`) names we have a single suffix say
`-tasks`  then if the filename has that it opens the file and looks for comments such as

```
%task: bibtex article1
%task: bibtex article2
```

and does each of those ? 

compared to the full set of tests there won't be so many needing this at all so I don't think the overhead here of looking inside  the files will be too great

mbertucci47:
Also, did you happen to see my comment about the `-BAD` tests in testfiles-compatible? The comment is somewhat hidden in one of your commits

mbertucci47:
@davidcarlisle There's an issue with the new config. The string `chapterbib-tasks-01` is hard-coded inside `runcmd` as the replacement for `%`

PR #1021, opened 2025-10-24
add bibtex or biber to test names

mbertucci47:
See #1020


mbertucci47:
Some xml files definitely still need updating. Sorry, I should have marked as a draft

PR #1024, opened 2025-10-25
fix some more tests

mbertucci47:
Work in progress. Added `mpost` and `mf` to allowable commands after `%tasks`.

### DIFF
--- a/tagging-status/testfiles-incompatible/feynmp-auto/feynmp-auto-tasks-01.pdftex.struct.xml
+++ b/tagging-status/testfiles-incompatible/feynmp-auto/feynmp-auto-tasks-01.pdftex.struct.xml
@@ -1,38 +1,50 @@
 <PDF>
  <StructTreeRoot>
   <Document xmlns="http://iso.org/pdf2/ssn"
-     id="ID.02"
+     id="ID.002"
     >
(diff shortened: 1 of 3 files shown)


davidcarlisle:
no objection to adding these although an alternative (as I did for the svg test, as I didn't want to install inkscape in the CI) is to check in the generated files, it's good to have the option to do either I think.

davidcarlisle:
@mbertucci47 
> @davidcarlisle As I've heard you are a master of languages, do you know of a Korean opentype font bundled with texlive? :-) The luatexko test is erroring because it uses Noto Serif KR which is not in texlive

당신은 올바르게 들었습니다

albatross couldn't find any. If you delete all the KR and use the generic Noto fonts the file runs without error but with a pile of missing character warnings, but you can still make the XML which shows the structure just with the text gone, that may be the best we can do here, I don't think we should start adding large CJK font sets to the CI (athough we could in theory)
the text isn't really needed to test the structure.

mbertucci47:
The feynmp-auto-01 pdftex test is failing but when I run `l3build save -c config-incompatible -e pdftex feynmp-auto-01` I get the same xml file, so I'm not sure what's happening there. It can be figured out tomorrow

Issue #874, opened 2025-06-04
marginalia package incompatible


mbertucci47:
It produces parent-child warnings. Example from the documentation:
```tex
\DocumentMetadata
  {
    lang=en-US,
    pdfversion=2.0,
    pdfstandard=ua-2,
    tagging=on,
  }
\documentclass{article}

\usepackage{marginalia}

\begin{document}

Here is some body text.\marginalia{Here is a marginal note.} Some more
body text.\marginalia[style=\footnotesize\itshape\raggedright]{Here is another
marginal note, set in smaller text and italics, whose position has been been
adjusted automatically.}
Some final body text.\marginalia[pos=left, valign=b, style=\sffamily\raggedleft,
width=35mm]{This note is placed on the left side of the page, wider, in sans
serif, ragged left, and bottom-aligned.}

\end{document}
```
warns e.g.
```
Package tagpdf Warning: Parent-Child 'pdf2:P' --> 'pdf2:Part'.
(tagpdf)                Relation is not allowed! on line 15
(tagpdf)                struct 6, text --> struct 7, text-unit
```

PR #1028, opened 2025-10-28
some test changes

mbertucci47:
Changes marginalia to compatible after discussion in #874.

Moves luamathalign-01 to incompatible-luatex (see #1029).

Removed `-BAD` from a luatexja test after package changes.

Removes the `testfiles` folder that just had a `.gitkeep` since I don't think it's needed any more.

Adds `-BAD` to abstract and typicons tests and adds an abstract test that produces a warning. For typicons and in general, should `-BAD` be added to any test that produces a PDF that is readable but not good? For example, all the symbol packages (pifont, dingbats, etc), all the math packages without luamml support, and all the packages that use lists for things that aren't lists (e.g. abstract)?

### DIFF
--- a/tagging-status/build.lua
+++ b/tagging-status/build.lua
@@ -2,7 +2,6 @@ module = "tagging-project-examples"
 
 test_order = {'structure_test'}
 includetests = {'*'}
-testfiledir=""
 checkengines = {'luatex', 'pdftex'}
 checkconfigs = {
   'config-compatible',


davidcarlisle:
> Hmm the `pdftex/luatex - build` tests are failing (I assume because I deleted the `testfiles` directory)

yes it wasn't entirely obvious how to say the top level configuration should enable checks but only in the directories of the included configurations. It's probably possible but leaving the empty `testfiles` there was easier at the time:-)

mbertucci47:
I will work on adding `-BAD` to tests in a future commit. I think this one can be merged now.

PR #1031, opened 2025-10-31
Rename bad tests with "-BAD"

mbertucci47:
Hopefully all self explanatory, though it wouldn't hurt to check. This would be a lot easier if one could see the full files in the github diff. My going through the tests was neither comprehensive nor systematic.

Also, small typo fixes, renamed the amsmath test to match the usual pattern, removed the local outdated copy of `asmeconf.cls`, and added a failing xcomment test.


davidcarlisle:
note the proof read edits to wtpdf/fulldoc look fine but would have been better as a separate PR (that's part of a separate project really)

tagging-status/testfiles-partial/asmeconf/asmeconf-01-template-BAD.pdftex.struct.xml
```diff
@@ -2487,15 +2488,15 @@
...
+       <?MarkedContent page="5" ?>3 team added support for PDF/A tothe
        <Span xmlns="http://iso.org/pdf2/ssn"
           id="ID.0572"
           actualtext="LaTeX"
           phoneme="[ˈlaːtɛx]"
          >
         <?MarkedContent page="5" ?>LATEX
        </Span>
-       <?MarkedContent page="5" ?>kernel, using the command \DocumentMetadata{..}.This approach works with both pdfLATEX and LuaLATEX. Notethat accessible conformance (a) is still under development by the
+       <?MarkedContent page="5" ?>kernel, using the command\DocumentMetadata{..}.This approach works with both pdfLATEX and LuaLATEX. Notethat accessible conformance (a) is still under development by the
```
davidcarlisle: this is a BAD file so we'll come back to it at some point but this seems to be a regression, we lost the space here?

davidcarlisle (approved):

tagging-status/testfiles-partial/asmeconf/asmeconf-01-template-BAD.pdftex.struct.xml
davidcarlisle: yes sorry about that I noticed later diff chunks had spaces coming and going due to line changes, that first one just caught my eye as it was just losing a space. But anyway it is what it is, we can't change pdftex output in this cycle.

PR #1049, opened 2025-11-07
add tasks for asymptote test

mbertucci47:
Also move to incompatible-mathml.


mbertucci47:
I can't reproduce the diff locally. Ideas are welcome.

davidcarlisle:
The problem is the usual libglut dance for setting up asy, the log shows

```
Transcript written on algorithm-01-BAD.log.
  asymptote-tasks-01 (3/27)
asy: error while loading shared libraries: libglut.so.3: cannot open shared object file: No such file or directory
asy: error while loading shared libraries: libglut.so.3: cannot open shared object file: No such file or directory
asy: error while loading shared libraries: libglut.so.3: cannot open shared object file: No such file or directory
```

I forget which ubuntu package libglut comes from but we could add it when setting up the vm or just consider the generated asy files as source and check them in rather than running asy in the CI

davidcarlisle:
can you rebase the workflow file to your branch and see if the tests pass, I have added libglut3(I think) and the test runs on main branch show it being installed via apt, but the runs here don't pick that up. We could merge anyway then clean up later if it doesn't work but testing on the branch would be better,

mbertucci47:
@davidcarlisle Yes I can but probably not until tomorrow. You are welcome to merge and fix if that’s easier

davidcarlisle:
so the test still fails with

```
asy: error while loading shared libraries: libglut.so.3: cannot open shared object file: No such file or directory
```

despite the new libglut step in the pipeline succeeding with

```
...
Setting up libglut-dev:amd64 (3.4.0-1build1) ...
Setting up freeglut3-dev:amd64 (3.4.0-1build1) ...

Running kernel seems to be up-to-date.

No services need to be restarted.

No containers need to be restarted.

No user sessions are running outdated binaries.
```

So I don't know, move that test to excluded for now?

PR #1063, opened 2025-11-14
update hologo status

mbertucci47:


mbertucci47:
Also updates the xml files for asmeconf and hvlogos which use hologo.

PR #1068, opened 2025-11-18
move luatexko to compatible

mbertucci47:
I changed the tests to just use latin letters since texlive doesn't include an opentype Korean font. I am certainly no expert but it seems the tagging looks alright, with all the TextDecoration attributes. It doesn't differentiate between the different types of strikeout or underline, but I don't know if that is possible like it is for CSS. Also not totally sure about the vertical text parts. Hopefully someone more knowledgeable can take a look.

### DIFF
--- a/tagging-status/testfiles-compatible-luatex/luatexko/luatexko-01.tex
+++ b/tagging-status/testfiles-compatible-luatex/luatexko/luatexko-01.tex
@@ -8,6 +8,7 @@
 \documentclass{article}
 
 \usepackage{unicode-math}
+\tagpdfsetup{role/user-NS=luatexko}% stable User NS
 \usepackage{luatexko}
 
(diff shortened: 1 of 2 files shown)


tagging-status/testfiles-compatible-luatex/luatexko/luatexko-01.tex
```diff
@@ -0,0 +1,78 @@
...
+  {
+    lang=ko,
+    pdfversion=2.0,
+    pdfstandard=ua-2,
+    tagging=on
+  }
+\documentclass{article}
+
+\usepackage{unicode-math}
+\usepackage{luatexko}
```
davidcarlisle:
```suggestion
\tagpdfsetup{role/user-NS=luatexko}% stable User NS
\usepackage{luatexko}
```

Issue #913, opened 2025-06-13
pstricks tagging keys don't seem to work


mbertucci47:
Here's a comparison with the latex-lab tikz keys and the keys provided from luamplib. All three seem to tag things differently (for example, `mplibcode` is a figure at the top level, while `tikzpicture` is inside a `<P>`) and I don't know which is correct, but pstricks definitely seems wrong since neither the alt text or actualtext show up in the tags.
```tex
\DocumentMetadata
  {
    lang=en-US,
    pdfversion=2.0,
    pdfstandard=ua-2,
    tagging=on,
  }
\documentclass{article}

\UseName{sys_if_engine_opentype:T}{\usepackage{unicode-math}}
\usepackage{tikz}
\usepackage{pstricks}
\usepackage{luamplib}

\begin{document}

\psset{unit=1cm}

\begin{tikzpicture}[alt=A square]
\draw (0,0) grid (1,1);
\end{tikzpicture}

\begin{pspicture}[alt=A diamond](0,0)(1,1)
\psdiamond(0,0)(1,1)
\end{pspicture}

\begin{mplibcode}[alt=alt text]
beginfig(1);
draw (0,0)--(20,0)--(20,20)--(0,20)--cycle;
endfig;
\end{mplibcode}

\begin{tikzpicture}[actualtext=a tikz test]
\draw (0,0) grid (1,1);
\end{tikzpicture}

\begin{pspicture}[actualtext=a pstricks test](0,0)(1,1)
\psdiamond(0,0)(1,1)
\end{pspicture}

\begin{mplibcode}[actualtext=just a test]
beginfig(1);
draw (0,0)--(20,0)--(20,20)--(0,20)--cycle;
endfig;
\end{mplibcode}

\begin{tikzpicture}[artifact]
\draw (0,0) grid (1,1);
\end{tikzpicture}

\begin{pspicture}[tag=artifact](0,0)(1,1)
\psdiamond(0,0)(1,1)
\end{pspicture}

\begin{mplibcode}[artifact]
beginfig(1);
draw (0,0)--(20,0)--(20,20)--(0,20)--cycle;
endfig;
\end{mplibcode}

\end{document}
```
```
Marked content on page -1:
Document (http://iso.org/pdf2/ssn):
├─Marked content on page -1:
├─text-unit (https://www.latex-project.org/ns/dflt) / Part (http://iso.org/pdf2/ssn):
│ ├─Marked content on page -1:
[...]

PR #1091, opened 2025-11-24
move pstricks to incompatible

mbertucci47:
See #913. Also marks all the pstricks-dependent packages as incompatible with a note.


davidcarlisle:
> Compatibility is only with lualatex; dvips cannot properly support PDF/UA.

I wonder if we should consider pstricks as luatex-only as it's incompatibility with pdftex isn't related to tagging.

mbertucci47:
It was already listed as luatex-only, but I also adjusted the comment to say "possible compatibility" since it is not currently compatible even with luatex.

Issue #1066, opened 2025-11-17
tcolorbox produces parent-child warnings


mbertucci47:
The example
```tex
\DocumentMetadata
  {
    lang=en-US,
    pdfversion=2.0,
    pdfstandard=ua-2,
    tagging=on,
  }
\documentclass{article}

\usepackage{tcolorbox}

\begin{document}

\begin{tcolorbox}
Some text
\end{tcolorbox}

\end{document}
```
warns
```
Package tagpdf Warning: Parent-Child 'pdf:StructTreeRoot' --> 'pdf2:Div'.
(tagpdf)                Relation is not allowed! on line 16
(tagpdf)                struct 1, StructTreeRoot --> struct 5, Div


Package tagpdf Warning: Parent-Child 'pdf:StructTreeRoot' --> 'pdf2:Part'.
(tagpdf)                Relation is not allowed! on line 16
(tagpdf)                struct 5, Div --> struct 6, text-unit


Package tagpdf Warning: Parent-Child 'pdf:StructTreeRoot' --> 'pdf2:P'.
(tagpdf)                Relation is not allowed! on line 16
(tagpdf)                struct 6, text-unit --> struct 7, text


Package tagpdf Warning: Parent-Child 'pdf:StructTreeRoot' --> 'pdf2:P'.
(tagpdf)                Relation is not allowed! on line 17
(tagpdf)                struct 1, StructTreeRoot --> struct 8, text


Package tagpdf Warning: Parent-Child 'pdf2:P' --> 'pdf2:Part'.
(tagpdf)                Relation is not allowed! on line 17
(tagpdf)                struct 8, text --> struct 9, text-unit


Package tagpdf Warning: Parent-Child 'pdf2:P' --> 'pdf2:P'.
(tagpdf)                Relation is not allowed! on line 17
(tagpdf)                struct 9, text-unit --> struct 10, text
```

PR #1098, opened 2025-12-02
Move tcolorbox to partially-compatible

mbertucci47:
See https://github.com/T-F-S/tcolorbox/issues/358 and #1066. Simple boxes are now tagged reasonably. I'm not sure how box titles should be tagged so I added a titled box to the test.


mbertucci47:
The same issue as in #1068 is appearing. I will add `\tagpdfsetup{role/user-NS=tcolorbox}` but maybe this should be fixed at the code level.

mbertucci47:
The other failures are due to a package change in systeme so I think this can be merged and those xml files updated later,

PR #1100, opened 2025-12-04
add asmejour tagging status

John02139:
This includes test files and additions to the yaml.  I think that some of additional test files will be generated from those included here, but me know if I need to add something.

Also in the yaml, minor edits to asmeconf and mitthesis status.

### DIFF
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -12513,10 +12513,20 @@
   luatex-only: true
   included-in:
   priority: 9
+  ctan-pkg: asmeconf
   comments: "Validates under luatex (don't use subcaptions)."
   issues: [1054]
   tests: true
-  updated: 2025-11-12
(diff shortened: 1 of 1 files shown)


davidcarlisle:
You haven't included the reference xml file (as made by l3build save)  the test works by running latex and show-pdf-tags --xml, and checking it gets the same xml as you did.

Also it would be better really if your test file used (say) `example-image-a` and `example-image-b` a couple of extra pdf images are not the end of the world, but if every package proovides two that's a lot of images we are carrying for not a lot of gain,

also the test file has to have `-bibtex` in its name to trigger the test runner to run bibtex or (as we did with asmeconf) you could check in the bbl rather than the bib so running bibtex isn't needed.

davidcarlisle:
@John02139 do you want to adjust or we could sort it out here?

PR #1136, opened 2025-12-15
emoji packages

mbertucci47:
Lists bxcoloremoji as partially-compatible because it uses twemojis which just inserts images without actualtext or tounicode values.

Lists emo as partially-compatible because it does the same thing with pdftex, and with luatex the symbols don't seem to appear in the PDF tags (not sure why this happens).

Lists emoji as compatible because the emojis are just used directly.


mbertucci47:
The failing test is because on linux, the emoji package uses the Noto Color Emoji fonts included in texlive, while on Windows it uses Segoe UI Emoji. For some reason the Noto fonts do not produce anything in show-pdf-tags but the Segoe fonts do. This is also why the symbols in the emo test don't appear with lualatex. @davidcarlisle Any ideas why the Noto fonts are a problem?

davidcarlisle:
> The failing test is because on linux, the emoji package uses the Noto Color Emoji fonts included in texlive, while on Windows it uses Segoe UI Emoji. For some reason the Noto fonts do not produce anything in show-pdf-tags but the Segoe fonts do. This is also why the symbols in the emo test don't appear with lualatex. @davidcarlisle Any ideas why the Noto fonts are a problem?

You can avoid platform dependencies by specifying the font (so long as you specify a font you have:-)

```
\usepackage{emoji}
%\setemojifont{NotoColorEmoji.ttf}
\setemojifont{Segoe UI Emoji}
```

I tested those two using a cygwin tex 

using Segoe UI Emoji the font is directly used in the PDF (and `show-pdf-tags` sees it as unicode text).

using NotoColorEmoji the font is not used at the PDF Level, the end of the log shows in the two cases:

```
</usr/local/texlive/2025/texmf-dist/fonts/opentype/public/lm/lmroman10-regular.o
tf></usr/share/fonts/microsoft/seguiemj.ttf>
Output written on emj.pdf (1 page, 44963 bytes).
```

and

```
<./luat
ex.W1VI3t/1.png><./luatex.W1VI3t/2.png><./luatex.W1VI3t/3.png><./luatex.W1VI3t/
4.png>]<<latex-list-css.html>><<latex-align-css.html>> (./emj.aux)
.....
</usr/local/texlive/2025/texmf-dist/fonts/opentype/public/lm/lmroman10-regular.
otf>
Output written on emj.pdf (1 page, 42622 bytes).
```


The Noto Color format isn't natively supported by PDF so luatex is extracting glyph images and inserting them.
They are not getting any `ActualText` so get lost to the  tagging.

As such I think `emoji` needs to be `partially-compatible` @zauguin  I don't know if luaotfload could add ActualText with the Uniocde text corresponding to the extracted glyphs?

mbertucci47:
So plan is use the Noto fonts, mark as `-BAD`, and move to partial-luatex, correct?

PR #1174, opened 2026-01-20
status update for prooftrees (v0.9.2 with forest-ext v0.2)

cfr42:
Current version auto-generates alt text for tableaux if tagging is active.

### DIFF
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -9470,14 +9470,14 @@
 
 - name: prooftrees
   type: package
-  status: currently-incompatible
+  status: compatible
   included-in:
   priority: 9
   issues: [1117]
-  tests: false
+  tests: true
   tasks: needs tests
-  comments: "Relies on incompatible forest package."
-  updated: 2025-12-11
+  comments: "Depends on forest-ext. Note tableaux are syntactic not semantic."
+  updated: 2026-01-19
 
 - name: psfrag
   type: package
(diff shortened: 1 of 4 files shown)


mbertucci47:
Is the math in the alt text useful? Doesn't it need mathml to be read correctly by a screen reader?

Also, not that it really matters, but why the odd number of spaces between things in the alt text?

davidcarlisle:
@cfr42  

>  The mml would be fine for a sighted reader, but would require somebody using a screen-reader to know that 'similar' should be understood as 'not', 

`<mo intent="not">~</mo>` would be read as "not", for this to happen of course the tex to mathml convertor needs a bit of help, but `\def\lnot{\MathMLintent{not}{\sim}}` would work. the braille would remain the single character representing `~` not the 5 characters representing backslash l n o t

davidcarlisle:
well as I say I don't think possible mathml is a blocker for this PR, what was blocking it though is that texlive hadn't updated (cache issues) so tests were failing, and having unblocked the texlive update, several other unrelated tests are now failing as things have been updated.
So I'll let the updates settle down and test suite passing before coming back to this PR, but personally I would be happy to merge this and make any small adjustments to yaml comments later

davidcarlisle:
test suite now passes again on main(unreleated to this PR), as discussed I'll merge this PR and add a comment to the yaml that tagging uses tex in alt (not saying taht is good or bad, just as a statement of fact)

PR #1176, opened 2026-01-20
direct pdf links

davidcarlisle:
we have checked in versions of the AF and SE tagged pdf, so can offer direct links to the pdf for testing in addition to linking to the online latex editor/texlive.net runner


FrankMittelbach (approved):
approved, with a question/suggestion

PR #1183, opened 2026-01-22
update eqnlines testfiles to v0.13

nbeisert:
updated testfiles for eqnlines (some interfaces changed) and corresponding status.

(unrelated: a hint in `CONTRIBUTING.md` regarding installing `show-pdf-tags` might help because `l3build` stalls otherwise without notice)


davidcarlisle:
> unrelated: a hint in CONTRIBUTING.md regarding installing show-pdf-tags might help because

Thanks, we can add something although it's in a default texlive install so we'd kind of assumed most people developing package tagging tests would have it already?

davidcarlisle:
apart from doc we could probably make l3build give a sensible error message if it's not there

davidcarlisle:
That's odd

This is rather off topic for the PR but..

I was going to suggest testing `mod`  so around line 44 of  `build.lua`

```
  local mod = loadfile(kpse.find_file('show-pdf-tags', 'lua'), 't', env)
  if mod == nil then error ("show-pdf-tags not found") end
````

but if I try that with `show-pdf-tags`  changed so it's not found, then `nil` isn't returned, instead the program hangs

@zauguin  is that expected?

it may not have hung forever it may be searching for the file but it took long enough that I killed the job.

Could instead check that the commandline version is in the path, something like

```
  local s = os.execute("show-pdf-tags -v")
  if s ~= 0 then error ("show-pdf-tags not found") end
  local mod = loadfile(kpse.find_file('show-pdf-tags', 'lua'), 't', env)
```

PR #1195, opened 2026-01-28
list xetex packages as no-support and ptex packages as incompatible

mbertucci47:
Preemptively lists several xetex- and ptex-specific packages as no-support.

### DIFF
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -5579,7 +5579,7 @@
 
 - name: fixjfm
   type: package
-  status: no-support
+  status: currently-incompatible
   included-in:
   priority: 9
   issues:
@@ -7452,7 +7452,7 @@
 - name: jslogo
   ctan-pkg: jsclasses
   type: package
-  status: no-support
+  status: currently-incompatible
   included-in:
   priority: 9
   issues:
@@ -7472,7 +7472,7 @@
 - name: jsverb
   ctan-pkg: jsclasses
   type: package
-  status: no-support
+  status: currently-incompatible
   included-in:
   priority: 9
   issues:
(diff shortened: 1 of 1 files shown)


davidcarlisle:
@mbertucci47 we were thinking that at least for the ptex ones it might be better to mark as incompatible rather than no-support but then add a new `ptex-only: true` field matching the existing `luatex-olny: true` that way we can still easily filter them out when needed

mbertucci47:
Should be good to go now. I used `ptex-only` regardless of whether a package works with ptex, uptex, or both.

mbertucci47:
Is it possible there's still an issue with texlive not updating? I'm getting a lot of failing tests in my own fork that are passing here.

FrankMittelbach:
> Should be good to go now. I used `ptex-only` regardless of whether a package works with ptex, uptex, or both.

I'm not sure this works @mbertucci47 . While ptex is not going to ever be fully compatible the situation with uptex is more that we don't have the knowledge to help much to solve any issues, but it might well be that those become compatible over time (or are already compatible).

mbertucci47:
@FrankMittelbach Right now tagpdf warns that uplatex does not support real space characters, which I thought was the main thing preventing xetex from being compatible. Also, aren't there issues with tagging and dvipdfmx? I vaguely remember something about this

davidcarlisle:
@FrankMittelbach  that's true but I'm not sure we can usefully distinguish here. Many packages need "some kind of ptex" and then have conditional code internally to work with the variant engines.  So I am not sure you can classify packages as uptex only or ptex-but-not-uptex 

 `pxjahyper.sty` for example having internal guards such as `\ifpxhy@uptex`

Issue #1197, opened 2026-01-30
-dev engines and tagpdf cannot process \index with "|" within headings


giovanni111:
lualatex-dev 2026.1.30 with
```
\DocumentMetadata{tagging=on}
\documentclass{article}
\usepackage{hyperref}
\makeindex
\begin{document}
\section{a\index{b|textbf}}
\end{document}
```
results in error, probably because of `\protected@edef\l__tag_sec_tmpa_tl {#4}` in the definition of `\__tag_set_title_hang:nNnn` in latex-lab-testphase-sec 2026-01-19 v0.84l.
Quick fix `{\let\index\@gobble\protected@xdef\l__tag_sec_tmpa_tl {#4}}` seems to work.

mbertucci47:
I think this has been fixed so I'll close here.

PR #1198, opened 2026-01-30
Adds tests and update package statuses

mbertucci47:
Lists a4, a4wide, advdate, anysize, cmap, mmap, as compatible without tests.

Adds a test for hyperref from #1197 that will error with the dev format.

Adds some tests for longtable.

Lists theoremref as compatible with tests.

Lists caption2 as incompatible with a test.

Lists appendixnumberbeamer and beamerposter as no-support.

Lists mathpartir and thumbs as partially-compatible with a comment and tests.

Changes the status of standalone to partially-compatible with a comment.


davidcarlisle:
ubuntu update is failing but unrelated to us, it seems, so merging here, it will hopefully fix itself

PR #1205, opened 2026-02-02
status updates

mbertucci47:
Adds several packages to tagging-status, most with a status but some unchecked. Also shortens a few long lines and lists the old presentation packages powerdot, prosper, and seminar as no-support since they require latex+dvips+ps2pdf.

### DIFF
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -4106,12 +4106,12 @@
 
 - name: cypriot
   type: package
-  status: unchecked
+  status: partially-compatible
   included-in:
   priority: 9
   issues:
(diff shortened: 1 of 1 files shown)


mbertucci47:
A _lot_ of failed tests... All unrelated to this PR. Haven't looked but my guess is the space issue that was fixed in tagpdf. I'll regenerate the xml files in a new PR.

mbertucci47:
Actually what was one MC is being broken up into many MC's. That doesn't look right to me. @u-fischer Do you know what's going on with the test diffs? I should also mention that it's only showing up with luatex

PR #1220, opened 2026-02-06
add and move tests

mbertucci47:
Adds a test for the tikz fadings library and moves multicolrule to partially-compatible.


davidcarlisle:
and the blank line

```
*** ./build/test-config-partial/tikzlibrary-fadings-01.struct.xml	2026-02-07 00:03:11.204375588 +0000
--- ./build/test-config-partial/tikzlibrary-fadings-01.luatex.pdf	2026-02-07 00:03:19.062397573 +0000
***************
*** 18,22 ****
     </text-unit>
    </Document>
   </StructTreeRoot>
- 
  </PDF>
--- 18,21 ----
```

means the xml was generated with the bad show-pdf-tags that broke all the tests yesterday (sorry:-)  it's failing now as the updated show-pdf-tags reached texlive

PR #1227, opened 2026-02-11
Status updates

mbertucci47:
Lists anyfontsize, arevtext, datenumber, docmute, emptypage, polski, rawfonts, savesym, and scrtime as compatible without tests.

Lists arevmath, pdfmysym, rsfso, and stix as partially compatible without a test.

Lists appendix and shuffle as partially compatible with tests.

Lists arabtex and utf8 as no support.

Lists collect as compatible with a test.

### DIFF
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -11176,16 +11176,6 @@
   tests: true
   updated: 2024-08-16
 
-- name: pdfmsym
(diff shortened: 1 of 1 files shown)


mbertucci47:
The eqnlines tests are failing but someone more knowledgeable of mathml will have to take a look at that.

PR #1242, opened 2026-02-25
Update testfiles rpgicons

jasperhabicht:
Future release (v2.6.0, released this weekend) will update use of `actualtext` key. New testfiles will raise error until new release is used. So, this PR should probably stay open until then.


jasperhabicht:
Version 2.6.0 uploaded to CTAN today. Checks on RP should succeed once the new files are installed.

PR #1247, opened 2026-02-26
add memoize-ext

cfr42:
I'm not sure if it would be better to just refuse to use memoize if tagging but I have documents which can't be realistically compiled at all in that case. Moreover, the tagging is actually more reliable for forest and prooftrees, in the sense that the bounding box is (I hope) correct if the are memoized, whereas it is merely approximated otherwise. For prooftrees, this is not so bad, but for forest it is more likely to be significantly wrong.

But, anyway, I am even less sure about this request ...

### DIFF
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -8548,8 +8548,8 @@
   package-repository: "https://codeberg.org/cfr/prooftrees"
   tests: true
   comments: "Adds sockets to the utilisation of externs created by Memoize.
-    actualtext/alt/artifact supported for TikZ pictures."
-  updated: 2026-02-26
+             actualtext/alt/artifact supported for TikZ pictures."
+  updated: 2026-03-20
 
 
 - name: mercatormap
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -8539,6 +8539,19 @@
   tasks: needs tests
   updated: 2024-07-18
 
+- name: memoize-ext
+  type: package
+  status: compatible
+  included-in:
+  priority: 9
+  issues: 
+  package-repository: "https://codeberg.org/cfr/prooftrees"
+  tests: true
+  comments: "Adds sockets to the utilisation of externs created by Memoize.
+    actualtext/alt/artifact supported for TikZ pictures."
+  updated: 2026-02-26
+
+
 - name: mercatormap
   type: package
   status: unchecked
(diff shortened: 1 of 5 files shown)


cfr42:
I realise that these tests will never pass with your configuration. So probably I should close this since it isn't really possible to write useful tests which will pass for your setup?

cfr42:
> The tagging looks reasonable to me but @u-fischer may want to take a look before merging.

As I said originally, I am really not sure about externalisation tools here. On the other hand, there are certainly cases where it seems unobjectionable.

cfr42:
@mbertucci47  @u-fischer @davidcarlisle  should I delete this? 

only I would like to know if there is something wrong with the tagging before I do that. because I do not much want to help somebody else do the same wrong thing, if wrong it is.

davidcarlisle:
personally I'd merge now, holding the PR open just creates work keeping the branches in sync (eg the current unrelated issue it's flagging with epic)  If issues come up with the tests later it wouldn't be the first package that has needed adjustment, but it's easier to test locally after a merge rather than having to pull your repo here

cfr42:
@davidcarlisle  OK, thanks. That is partly why I was going to delete it.

cfr42:
@davidcarlisle  any objection if I just delete this? there does not seem much point in updating it, frankly. stuff which uses it is marked as compatible, even if this isn't, and I think that will have to do.

I hoped it would either get approved or rejected because that would be useful in terms of what might or might not work for `memoize`, but I don't want to try tangling with `git` following a 2e release ;).

davidcarlisle:
@cfr42  can't we simply merge, there are no git conflicts and the test failures are unrelated and being addressed elsewhere, But if you want to delete i don't object as such but I'm not sure why you want to delete? If you delete it just stays as "unknown" and will come up later.

cfr42:
> @cfr42 can't we simply merge, there are no git conflicts and the test failures are unrelated and being addressed elsewhere, But if you want to delete i don't object as such but I'm not sure why you want to delete? If you delete it just stays as "unknown" and will come up later.

I am happy if you merge it. I want to delete it because I don't think that is at all likely to happen and I would rather be done with it in that case rather than having it in limbo. (I was going to say 'perpetual limbo' but I am pretty sure that limbo was perpetual by definition, until they abolished it.) And I am afraid that somebody will ask me to update the branch again at some point, which I suspect will far exceed my limited `git` skills.

cfr42:
Sorry, but I do not see the point.

PR #1248, opened 2026-02-26
update tests

mbertucci47:
Updates some test files. @cfr42 Is it correct that the adforn characters become blank spaces? If everything looks good I think those packages could be moved to compatible.


cfr42:
> @cfr42 I changed them to compatible and made a comment about needing to be loaded via the package.

Thanks very much. This was on my todo list ... ;).

PR #1265, opened 2026-03-12
Use lua-unicode-math in tests instead of unicode-math

mbertucci47:
(This PR is not meant to be merged, at least not as-is. I also opened an issue to track this; see #1266.)

For the purposes of experimentation and speeding up the test suite, I tried changing every instance of `\usepackage{unicode-math}` in the test files with `\usepackage{lua-unicode-math}\tagpdfsetup{math/mathml/luamml/load}` and, where necessary, `\usepackage{fontspec}`. Some tests had no change, but many did have changes in the resulting mathml. I'm opening this PR just to provide a simple way to view the diffs of the xml files. Some changes may be expected, for example the several tests that error because of the different timing of when lua-unicode-math vs. unicode-math defines the symbols (immediately vs. at begindocument). I adjusted a few tests to avoid this but others produce confusing errors like MnSymbol and fdsymbol. I am hoping this will be useful to those who can look at the diffs and decide if the changes are expected or not, so @zauguin and @davidcarlisle I'd guess.

A few other observations:
- The prime character `′` in mathml goes from `<mo lspace="0" rspace="0">′</mo>` to `<mi mathvariant="normal">′</mi>`. Same for `!`.
- Sometimes several MCs get collapsed into one, sometimes the other way around.
- Since all the mathml is put on one line, some of the diffs are very difficult to parse. I wonder how difficult it would be to provide better formatting of the mathml. See, for one example, amsmath-01.
- I did not include any changes that are due to changes in luatex and luamml (see comments in #1254).
- In at least one case an entire `AssociatedFile` disappears; see numerica-01. In another an entire `Formula` disappears; see longdivision-01.
- In some cases characters change, e.g. extarrows-01.
- In letterswitharrows-01, `<mover> <mi mathvariant="normal">x</mi> <mo stretchy="false">→</mo> </mover>` becomes `<mi> <mglyph/> </mi>` which definitely seems wrong.


zauguin:
> * The prime character `′` in mathml goes from `<mo lspace="0" rspace="0">′</mo>` to `<mi mathvariant="normal">′</mi>`. Same for `!`.

For `′` this is a luamml bug which is fixed in `main`. For `!` this is a lua-unicode-math bug which will be fixed with the next release.



> * Since all the mathml is put on one line, some of the diffs are very difficult to parse. I wonder how difficult it would be to provide better formatting of the mathml. See, for one example, amsmath-01.

`show-pdf-tags` just preserves whatever formatting the PDF attachment does or doesn't have and `luamml` can produce formatted output, but the whole process of writing it to a file and reading the file again strips formatting. I'm not sure if it's easier to preserve formatting there or to run something like `xmltidy` on the `show-pdf-tags` output.

> * In letterswitharrows-01, `<mover> <mi mathvariant="normal">x</mi> <mo stretchy="false">→</mo> </mover>` becomes `<mi> <mglyph/> </mi>` which definitely seems wrong.

That's actually somewhat correct. `letterswitharrows` overwrites `\vec` when loaded with the `vec-cev` option with a non-tagging compatible version. Since `unicode-math` overwrites everything at `\begin{document}` it restores the compatible `\vec`, while `lua-unicode-math` keeps the overwritten one. Given that the point of that option is to get another `\vec` I would consider keeping it to be the correct behavior.

mbertucci47:
@zauguin Are there plans for a luamml and lua-unicode-math release to fix the `'` and `!` issues?

PR #1278, opened 2026-03-23
add closed-issues and move epic to compatible

mbertucci47:
Adds `closed-issues` to the YAML file and moves epic to compatible. I don't know if the HTML conversion will just ignore the unknown type or if something needs to be done there.


FrankMittelbach:
the strikethrough is not really visible for me with my current browser settings, but it works well enough given the intro "Closed:"

Issue #1314, opened 2026-04-15
Table of contents break with ocgx2's ocgcolorlinks


JesseStraat:
In the following MWE,
```LaTeX
\DocumentMetadata{
    lang        = en,
    pdfstandard = ua-2,
    tagging=on,
    tagging-setup={math/setup=mathml-SE} 
}
\documentclass{article}

\usepackage{hyperref}
\usepackage[ocgcolorlinks]{ocgx2}

\begin{document}
\tableofcontents
%
\section{foo}
%
\section*{bar}
\addcontentsline{toc}{section}{bar}
\end{document}
```
the first section appears invisible in the table of content. However, `\addcontentsline` works as expected.

u-fischer:
The problem is the `\numberline` that typesets the number in the table of contents.  With tagging this is marked up as a Lbl and in the PDF stream the changes of the text operator (`Tr`) done by ocgx2 get separated by the code from the tagging and so are no longer correctly nested. Imho this must be discussed with the ocgx maintainer. So please report it there.

The only work-around that I can offer for now is to avoid the tagging of the number. Be aware that the problem can appear in other places too if links contain additional tagging, e.g. `\href{https://latex-project.org}{some text with \emph{tagging} }`, so I would say that ocgx2 can only be used with great care. 

~~~~
\DocumentMetadata{
    lang        = en,
    pdfstandard = ua-2,
    tagging=on,
    tagging-setup={math/setup=mathml-SE},    
    %uncompress
}
\documentclass{article}
 
\usepackage{hyperref}
\usepackage[ocgcolorlinks]{ocgx2}

\RemoveFromHook{contentsline/number/before}[tagpdf]
\RemoveFromHook{contentsline/number/after}[tagpdf]

\AddToHook{contentsline/number/after}{\pdffakespace}

   
\begin{document}

\tableofcontents

\addcontentsline{toc}{section}{\protect\numberline{1}bar}
\addcontentsline{toc}{section}{bar}

\addcontentsline{toc}{subsection}{\protect\numberline{1}bar}
\addcontentsline{toc}{subsection}{bar}

\end{document}
~~~~

FrankMittelbach:
So I guess this should be listed as incompatible or partially-compatible in the next update of the status table. @mbertucci47 ?

PR #1315, opened 2026-04-15
update ocgx2 status

mbertucci47:
Lists ocgx2 as incompatible. See #1314. Also links a few new issues


mbertucci47:
The wrapstuff tests are erroring only with pdftex. I just checked and they work fine without tagging and with luatex+tagging. @u-fischer did something change that might cause that? The package hasn't been updated so it must be something else

PR #1329, opened 2026-04-30
Update rpgicons testfiles

jasperhabicht:
Some new icons added and tagging mechanisms improved in the PGF variant. Needs version 2.9.0 of the package which is already on CTAN.


jasperhabicht:
The checks should pass once version 2.9.0 of the rpgicons package is installed.

PR #1336, opened 2026-05-11
Update dev tests

mbertucci47:
This updates the xml files for the tests I'm fairly confident are improved or innocuously changed. After I finish, anything failing is something I'll need someone else to check. I updated the workflow to use the dev formats in l3build.


davidcarlisle:
Thanks for these, we'll merge and see what's left....

PR #1339, opened 2026-05-11
Status updates

mbertucci47:
Moves tocbibind, anonchap, and secdot to partially compatible. More updates on the way.


mbertucci47:
Okay should be ready for review now. This moves
- colortbl and numerica-tables to compatible;
- anonchap, secdot, and tocbibind to partially-compatible;
- and newproof, smartref, theoremref, and titleref to currently-incompatible.

It also updates several other xml files.

Issue #1346, opened 2026-05-11
titleref functionality broken with recent dev release


mbertucci47:
The command `\titleref` produces an empty reference. Note the package was already incompatible with hyperref.
```tex
\DocumentMetadata
  {
    lang=en-US,
    pdfversion=2.0,
    pdfstandard=ua-2,
    tagging=on
  }
\documentclass{article}

\usepackage{titleref}

\title{titleref tagging test}

\begin{document}

\tableofcontents

\section[short]{Some section}
\label{sec}

Some text.

\section{Another section}

See \titleref{sec}.

\section*{An unnumbered section}
\label{unnum}

\titleref{unnum}

\end{document}
```

u-fischer:
I have a tendency to set that as "no-support".

FrankMittelbach:
Do we have a recomendation what to use instead? If so we should document that and set it to no-support.

u-fischer:
nameref.  `\usepackage{nameref}\let\titleref\nameref` should simply work as a replacement.

mbertucci47:
Moved to no-support in #1349

davidcarlisle:
we could ask if `titleref.sty` could be changed to be
```
\RequirePackage{nameref}
\let\titleref\nameref
```

then it would be compatible with hyperref and tagging

u-fischer:
For a real replacement a bit more is needed. I would suggest this here (which additionally adds a starred version for unlinked references:

~~~~latex
\documentclass{article}

\usepackage{hyperref}
\makeatletter
%%%% new titleref
\usepackage{nameref}
\newcommand{\theTitleReference}[2]{#2}
\DeclareRobustCommand{\currenttitle}{\@currentlabelname}
\NewDocumentCommand\titleref{sm}{%
 \IfBooleanTF{#1}
  {\theTitleReference{{\ref*{#2}}}{{\nameref*{#2}}}}
  {\theTitleReference{{\ref{#2}}}{{\nameref{#2}}}}}
%%%% end new titleref
\makeatother
% from docu
\renewcommand{\theTitleReference}[2]{#1\ \emph#2}
\begin{document}
\section[short]{Some section}
\label{sec}

Some text.

\section{Another section} \currenttitle

See \titleref{sec} and \titleref*{sec}.

\section*{An unnumbered section} \currenttitle
\label{unnum}

\titleref{unnum}
\end{document}
~~~~

PR #1349, opened 2026-05-14
move titleref to no-support

mbertucci47:
See the comments in #1346


mbertucci47:
See comments in matrix

PR #1359, opened 2026-05-30
Merge dev tests brance after June release

mbertucci47:
I'm sure some tests will be failing at first. Will fix as needed.


mbertucci47:
Is there a way to make it ignore the test.yaml change so it doesn't make all the tests use the dev format? It can also just be changed after merging.

davidcarlisle:
just change the yaml back after the merge I think.
for one-off test runs using -dev format you don't need to change the yaml it should be settable from the UI setting the  option on the workflow dispatch

mbertucci47:
I updated some tests with irrelevant changes like losing a space at the end of a line. All other failing tests need to be looked at individually. I think this can be merged now and the other tests sorted separately.

u-fischer (approved):

Issue #649, opened 2024-08-22
minitoc package incompatible


mbertucci47:
`\minitoc` of the [minitoc](https://ctan.org/pkg/minitoc) package produces parent-child warnings.
```tex
\DocumentMetadata
  {
    lang=en-US,
    pdfversion=2.0,
    pdfstandard=ua-2,
    testphase={phase-III,math,title,table,firstaid}
  }
\documentclass{report}
\usepackage{minitoc}
\usepackage{kantlipsum}

\title{minitoc tagging test}

\begin{document}
\dominitoc
\tableofcontents
\chapter{Primum capitulum}
\minitoc
\section{Prima sectio}
\kant[1-2]
\section{Secunda sectio}
\kant[3-4]
\chapter{Secundum capitulum}
\minitoc
\section{Prima sectio}
\kant[5-6]
\section{Secunda sectio}
\kant[7-8]
\section{Tertia sectio}
\kant[9-10]
\end{document}
```

mbertucci47:
The example no longer produces parent-child warnings and the tagging seems okay.

u-fischer:
But the output is wrong: there are no minitocs at all.

u-fischer:
minitoc patches heading commands (mostly to step counters to keep track of the minitoc content) and this no longer works with the templates. The minitocs for the example reappear if one step the counters manually (but tagging is still wrong):
The stepping is done imho not for starred chapters. 

~~~~
\DocumentMetadata
  {
    lang=en-US,
    pdfversion=2.0,
    pdfstandard=ua-2,
    tagging=on
  }
\documentclass{report}
\usepackage{minitoc}
\usepackage{kantlipsum}

\title{minitoc tagging test}

\begin{document}
\dominitoc
\tableofcontents

\stepcounter{mtc} %<----
\chapter{Primum capitulum}
\minitoc
\section{Prima sectio}
\kant[1-2]
\section{Secunda sectio}
\kant[3-4]

\stepcounter{mtc} %<-----
\chapter{Secundum capitulum}
\minitoc
\section{Prima sectio}
\kant[5-6]
\section{Secunda sectio}
\kant[7-8]
\section{Tertia sectio}
\kant[9-10]
\end{document} 
~~~~

PR #1369, opened 2026-06-06
update more tests

mbertucci47:
These test updates aren't necessarily expected but don't change the status of the involved packages.
- zref-clever: Two `<mtd> <mspace width="-9.963pt"/> </mtd>` were added in the mathml. Not really sure what that's about.
- fancypar: no longer errors but the tagging is wrong
- kanbun: no longer errors but the tagging is wrong
- minitoc: the minitocs have totally disappeared. No idea what's happening here


u-fischer:
I added a comment about the minitoc problem at https://github.com/latex3/tagging-project/issues/649, fancypar probably runs now because I fixed the tikz key issue.

u-fischer (approved):

Issue #1403, opened 2026-06-11
amsthm+thmtools styling incompatibility


JesseStraat:
The `\declaretheoremstyle` command from `thmtools` doesn't work if the used backend is `amsthm`. MWE:

```LaTeX
\DocumentMetadata{}
\documentclass{article}
\usepackage{amsthm}
\usepackage{thmtools}

\declaretheoremstyle[bodyfont=\slshape]{foobar}
\theoremstyle{foobar}
\declaretheorem{theorem}

\begin{document}
\begin{theorem}
    a
\end{theorem}
\end{document}
```
This leads to the warning 
```
Package thmtools Warning: Your backend doesn't allow styling theorems on input 
line 5.
```
Putting it directly into the `\declaretheorem` as `\declaretheorem[style=foobar]{theorem}` gives the error
```
! Package thmtools Error: \declaretheorem key `bodyfont' not known.
```

If `ntheorem` is used instead, `\declaretheoremstyle` is replaced with `\newtheoremstyle`, or `\DocumentMetadata{}` is removed, the issue is fixed, and results are as expected.

FrankMittelbach:
The `thmtools` package is already listed as currently incompatible, so this is just another aspect of it.

mbertucci47:
@FrankMittelbach The problem comes down to the fact that `\IfPackageLoadedTF{amsthm}` takes the false branch when `\DocumentMetadata` is loaded since the tagging code suppresses the actual loading of amsthm.sty.
```tex
\DocumentMetadata{}
\documentclass{article}
\usepackage{amsthm}
\begin{document}
\IfPackageLoadedTF{amsthm}{TRUE}{FALSE}
\end{document}
```
In thmtools.sty there is
```tex
\@ifpackageloaded{amsthm}{%
  \RequirePackage{thm-amsthm}
}{...}
```

FrankMittelbach:
@mbertucci47 true. That part can be resolved and I will do that, but eventually the different backend interfaces should be adjusted as well.

Issue #1364, opened 2026-06-02
enumext test erroring


mbertucci47:
The test enumext-01 is erroring with the June release. The other tests seem okay.
```tex
\DocumentMetadata
  {
    lang=en-US,
    pdfversion=2.0,
    pdfstandard=ua-2,
    tagging=on,
    %uncompress,
  }
\documentclass{article}
\usepackage[top=2cm,bottom=2cm,left=1cm,right=1cm]{geometry}
\usepackage{hyperref,enumext}
\hypersetup
  {
    colorlinks = true,
    pdftitle   = {Test tagged PDF for enumext package},
  }
\begin{document}
\section{only enumext}
Some text here

\begin{enumext}[labelwidth=1cm,labelsep=0.25cm,align=left, wrap-label=\textbf{(#1}, label=\arabic*.)]
  \item This text is in the first level.\label{A}
    \begin{enumext}
      \item This text is in the second level.\footnote{foot note in enumext}
        \begin{enumext}
          \item This text is in the third level.
            \begin{enumext}
              \item This text is in the fourth level.\footnotemark[20]
            \end{enumext}
        \end{enumext}
    \end{enumext}
  \item[X] This text is in the first level.
  \item* This text is in the first level.
\end{enumext}

Some ref here \ref{A}\footnotetext[20]{foot note in enumext [20]}

\section{only enumext*}

Some text here

\begin{enumext*}[columns=2,labelwidth=1cm,labelsep=0.25cm,align=left, wrap-label=\textbf{#1)}, label=(\arabic*.]
  \item This text is in the first level.\label{B}
  \item[X] This text is in the first level.\footnotemark[30]
  \item* This text is in the first level.\footnote{foot note in enumext*}
  \item This text is in the first level.
\end{enumext*}

Some ref here \ref{B}\footnotetext[30]{foot note in enumext* [30]}

\section{enumext nested in enumext*}

Some text here

\begin{enumext*}
  \item This text is in the first level.\label{C}
  \item \begin{enumext}[base-fix]
              \item This text is in nested enumext.\label{D}\footnote{foot note in neseted enumext}
              \item This text is in nested enumext.\footnotemark[40]
         \end{enumext}
  \item This text is in the first level.
\end{enumext*}

Some refs here \ref{C} and \ref{D} \footnotetext[40]{foot note in neseted enumext [40]}

\section{enumext* nested in enumext}

\begin{enumext}
  \item This text is in the first level. \label{E}
[...]

Issue #1342, opened 2026-05-11
error if no space after `filecontentshere` environment


mbertucci47:
Extracted from the test file.
```tex
\DocumentMetadata
  {
    lang=en-US,
    pdfversion=2.0,
    pdfstandard=ua-2,
    tagging=on
  }
\documentclass{article}

\usepackage{filecontentsdef}

\title{filecontentsdef tagging test}

\newtheorem{theorem}{Theorem}

\begin{document}

\begin{filecontentshere}{mytest.txt}
\begin{theorem}
Some theorem text
\end{theorem}
\end{filecontentshere}
\filecontentsexec\filecontentsheremacro

\end{document}
```
errors with
```
! Package tagpdf Error: The number of automatic begin (2) and end (1)
(tagpdf)                text-unit para hooks differ!
```
There's no error if a blank line is added after `\end{filecontentshere}`.

Issue #1337, opened 2026-05-11
\footnotetext  after list gives tagging error


u-fischer:
~~~~latex
\DocumentMetadata{tagging=on}
\documentclass{article}
\begin{document}
\begin{itemize}
\item abc
\end{itemize}
\footnotetext[1]{blub}
some text
\end{document}
~~~~

errors (in main + dev) with 

~~~~
! Package tagpdf Error: The number of automatic begin (2) and end (3)
(tagpdf)                text-unit para hooks differ!
~~~~

the problem is the ` \par \color@endgroup`  at the end of the `\fnote_footnotetext:n`. If `\@endpe` is true, both the `\par` and the `\endgraf`  insert structure-ends.

Switching endpe avoids the error, but the tagging structure is not fully correct, as the following text is in a new semantic paragraph:

~~~~latex
\DocumentMetadata{tagging=on}
\documentclass{article}
\ExplSyntaxOn\makeatletter 
\cs_set_protected:Npn \fnote_footnotetext:n #1 {
  \__fnote_debug_footnotetext:
  \hook_use:n {fntext/before}
  \tag_socket_use:n {fntext/begin}
  \socket_use:nn {fntext/process}
    {
    \hook_use:n {fntext}
    \reset@font
    \footnotesize
    \interlinepenalty\interfootnotelinepenalty
    \splittopskip\footnotesep
    \splitmaxdepth \dp\strutbox
    \floatingpenalty \@MM
    \hsize\columnwidth
    \@parboxrestore
    \UseTaggingSocket{para/restore}
    \parindent 1em     % typical default used in \@makefntext moved up here
    \def\@currentcounter{footnote}
    \protected@edef \@currentlabel { \p@footnote \@thefnmark }
    \hook_use:n {fntext/para}
    \color@begingroup
      \socket_use:nn {fntext/make}
        {
          \socket_use:n {fntext/begin}%
          \hook_use:n {fntext/begin}
          \ignorespaces
          #1
          \hook_use:n {fntext/end}
          \socket_use:n {fntext/end}
        }      
      \par
      \@endpefalse%<-----------
      \color@endgroup
  }
  \tag_socket_use:n{fntext/end}
  \hook_use:n {fntext/after}
}
\ExplSyntaxOff
\begin{document}
\begin{itemize}
\item abc
\end{itemize}
\footnotetext[1]{blub}
abc
\end{document}
~~~~

[...]

PR #1414, opened 2026-06-13
update more tests and comments on remaining failing tests

mbertucci47:
Updates some remaining tests.
- thmtools: the test errors because of https://github.com/muzimuzhi/thmtools/issues/75 which is unrelated to tagging so I just updated the xml with the error. Also added a test for https://github.com/latex3/tagging-project/issues/1403.
- eqnlines: no unexpected changes except the additional `<mspace width="-9.963pt"/>` which has shown up in other tests
- ghsystem: the angle brackets are now being tagged as math since they use `\langle` and `\rangle`. This seems wrong to me (see my comments in https://github.com/latex3/tagging-project/issues/1075) but I guess there's nothing one can do about it.

This leaves the following tests still failing:
- filecontentsdef-01: see https://github.com/latex3/tagging-project/issues/1342
- zref-clever-03: this is unrelated to tagging and would be fixed by the code in https://github.com/latex3/latex2e/issues/2100 if added to firstaid
- enumext-01: see https://github.com/latex3/tagging-project/issues/1364, which is really just https://github.com/latex3/tagging-project/issues/1337
- letterswitharrows-01: see https://github.com/max-te/letterswitharrows/issues/1


FrankMittelbach:
> * [](https://github.com/latex3/tagging-project/issues/1337)
> 
> * letterswitharrows-01: see [faulty output with LaTeX 2026-06-01 max-te/letterswitharrows#1](https://github.com/max-te/letterswitharrows/issues/1)

the last one is generally broken by the latest release (tagging or not). Reason is somewhat strange coding using internals that changed. @u-fischer opened an issue at its repo

mbertucci47:
@FrankMittelbach Right, that's the issue I linked to

davidcarlisle:
if the package doesn't seem like it will get a quick update we could move the tests to `testfiles-excluded` so it doesn't don't show as failures every run,

FrankMittelbach:
> if the package doesn't seem like it will get a quick update we could move the tests to `testfiles-excluded` so it doesn't don't show as failures every run,

we could also check in the faulty result then we see it changing when it updates

mbertucci47:
I can update the test for letterswitharrows and add `-BAD` until it gets fixed. Should I do the same for zref-clever and enumext? Since presumably those will eventually get fixed.

As for filecontentsdef, I do not know why that is erroring.

u-fischer:
Didn't you already report the filecontentsdef error: https://github.com/latex3/tagging-project/issues/1342

mbertucci47:
@u-fischer Yes, I just mean I don't know why it's happening (and did not happen before)

mbertucci47:
I think this is good to merge now

PR #1424, opened 2026-06-17
Adding test files for `hawkdraw`

jasperhabicht:


mbertucci47:
hawkdraw-01.tex needs to be in the hawkdraw subfolder of testfiles-unchecked, otherwise looks good to me

jasperhabicht:
I wonder whether I can already place this into `testfiles-conpatible`. I thought that somebody could check the structure maybe. It looks fine to me, but I will do a check of the resulting PDF again.

mbertucci47:
Can you add a tagging-status.yml entry?

jasperhabicht:
> Can you add a tagging-status.yml entry?

I guess I forgot to add a `priority`, unsure whether it is required or not.

FrankMittelbach:
> > 
> 
> I guess I forgot to add a `priority`, unsure whether it is required or not.

we aren't using that really, that's a recent addition by github we still figure out if that is of any use

mbertucci47:
@jasperhabicht Added in https://github.com/latex3/tagging-project/commit/0871b9d7227b7e644198e96364d59e597094b35e

jasperhabicht:
Could it be that if no `priority` is given, the relevant entry does not appear on the list on https://latex3.github.io/tagging-project/tagging-status/ ? 

`hawkdraw` was not in the list, so I checked the YAML and saw the missing `priority`. Now, it appears in the list.

davidcarlisle:
@FrankMittelbach,  @jasperhabicht  meant our yaml priority field not a native github feature.  The entry would have shown in the full list at
https://latex3.github.io/tagging-project/tagging-status/full

but the top level tagging-status page filters some things including

```
{% if p.priority < 10 %}
```

at one point I think that was set lower but with <10 it is essentially filtering out entries with no priority.

FrankMittelbach:
Yeah my bad. I picked up the wrong context and was talking about the newly introduced priority field on GitHub which we don't really know what to do with yet if anything.

jasperhabicht:
All good! The issue was resolved and my question was answered. Thank you all for your support!

PR #1439, opened 2026-06-22
Update the amsldoc examples

u-fischer:


project-examples/amsldoc/amsldoc-tagged.tex
```diff
@@ -1123,12 +1128,19 @@ \section{Interrupting a display}
 it up again afterwards. \cn{intertext} may only appear right after a
 \cn{\\} or \cn{\\*} command. Notice the position of the word \qq{and} in
 this example.
+% tagging ========
+\iftaggingAF
+\\\textbf{Suppressed as tagging with MathML-AF incorrect}\\
+\url{https://github.com/latex3/tagging-project/issues/1033}
```
davidcarlisle, line 1134: 1033 is about `split` in equation?

davidcarlisle (changes requested):
I don't think we should check in generated files (ilg, toc, idx ...)

one query about the split issue made inline

project-examples/amsldoc/amsldoc-luamml.tex
```diff
@@ -1,2910 +1,16 @@
 % !Mode:: "TeX:UTF-8:Main"
-%
-% Copyright (C) 1995, 2000, 2013 American Mathematical Society.
```
davidcarlisle, line 3: we should probably keep these copyright statements

u-fischer: ? That is only a wrapper file which loads the other file.

project-examples/amsldoc/amsldoc-tagged.tex
u-fischer: The check-in of the generated files was simply an error, but I removed them already, 1033 is about split in equation.

project-examples/amsldoc/amsldoc-luamml.tex
davidcarlisle: sorry so it is (now:-) it wasn't before and I confused myself looking at the diff, ignore this comment then.

FrankMittelbach (commented):
I agree with David that we shouldn't have .idx and ind files as part of the repo. when/why did they get added?

If we drop a few bits because they aren't yet compatible with tagging we should perhaps say so at the very top in the abstract or whereever

FrankMittelbach:
> I agree with David that we shouldn't have .idx and ind files as part of the repo. when/why did they get added?

ah, see that got already address ... wrong order of reading

davidcarlisle:
filecontentsdef is failing, but I think I'll merge and sort out locally

Issue #210, opened 2024-07-19
lineno


u-fischer:

u-fischer:
It looses line numbers and line numbers are not correctly tagged (as artifct or something else).

~~~~
\DocumentMetadata
  {
    lang=en-US,
    pdfversion=2.0,
    pdfstandard=ua-2,
    tagging=on,
    uncompress
  }
\documentclass{article}
\usepackage{lineno}
\usepackage{lipsum}

\linenumbers
\begin{document}

\section{Some heading}

bla\lipsum[1]

\end{document}
~~~~

gives

![image](https://github.com/user-attachments/assets/68db0ab0-f0a6-4086-bd96-e37fa7516c62)

mbertucci47:
See also #162

Udi-Fogiel:
I know next to nothing about tagging, but I just released the [lualineno](https://ctan.org/pkg/lualineno) package, which seems to work 
```tex
\DocumentMetadata
  {
    lang=en-US,
    pdfversion=2.0,
    pdfstandard=ua-2,
    tagging=on,
    uncompress
  }
\documentclass{article}
\usepackage{lualineno}
\newcounter{lineno}
\lualineno{
    define = {
        name = foo
        toks = {\stepcounter{lineno}}
        left = {\tiny\thelineno\kern.8em}
    }
    set=foo
}

\usepackage{lipsum}

\begin{document}

\section{Some heading}

bla\lipsum[1]

\end{document}
```
but I'm not sure if the numbers are tagged nor how to check that.  If there are issues in this regard I'll happily to fix them (or at least try).

davidcarlisle:
Thanks for the file.

You can see the tags by using

```
show-pdf-tags --xml youfile.pdf
```

(`show-pdf-tags` being a texlua script in texlive by default)

If I add a title so

```
\DocumentMetadata
  {
    lang=en-US,
    pdfversion=2.0,
    pdfstandard=ua-2,
    tagging=on,
    uncompress
  }
\documentclass{article}
\usepackage{lualineno}
\newcounter{lineno}
\lualineno{
    define = {
        name = foo
        toks = {\stepcounter{lineno}}
        left = {\tiny\thelineno\kern.8em}
    }
    set=foo
}

\usepackage{lipsum}

\begin{document}

\title{test}
\maketitle

\section{Some heading}

bla\lipsum[1]

\end{document}
```

[...]

PR #1456, opened 2026-06-29
List lualineno as compatible

mbertucci47:
The line numbers are tagged as artifacts unless they contain content with structure, in which case tagging must be suspended. This is documented by the package and a comment is added in the yml file. I also added a comment in the lineno entry suggesting lualineno as a compatible alternative. See also the discussion in #210.

### DIFF
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -8195,8 +8195,8 @@
   status: compatible
   priority: 9
   luatex-only: true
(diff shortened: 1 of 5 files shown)


mbertucci47:
@u-fischer Does this structure tree look right to you?

mbertucci47:
I'll go ahead and merge this because it looks good to me. We can revert if there are any issues.

PR #1461, opened 2026-06-30
split partial-mathml test runs in two

mbertucci47:
The partial-mathml tests take the longest of the chunks currently so I've split them into two chunks.


FrankMittelbach (approved):
would be nice if there would be an easy way to say split this set in half, but given that this isn't the case ...

PR #1495, opened 2026-07-23
move kbordermatrix tests from excluded

mbertucci47:
Since it was added to texlive.

Also links some more repos.


mbertucci47:
Some of CI runs are failing because they fail to restore texlive from cache. Not sure what that's about but all the tests pass locally so I'll merge here.

PR #1505, opened 2026-07-27
update tests with cursive fonts

mbertucci47:
I guess these changed due to the tagpdf update. @u-fischer ? Doesn't matter too much since both are already incompatible


u-fischer:
Yes that is from the tagpdf update. Due to the improved handling of hyphens from hyphenation one gets now more marked content blocks. (This can happen in "GOOD" examples too ...)

PR #1513, opened 2026-08-06
status updates

mbertucci47:
Lists ducksay and testhyphens as incompatible with tests.

Lists enverb and luadraw as partially compatible with tests.

Lists fixltxhyph as compatible without a test.


mbertucci47:
Hmm the PR didn't trigger the CI run. Not sure why

FrankMittelbach:
seems that github is behaving oddly today, not the only issue even if they claim everything is functional

Issue #1469, opened 2026-07-01
tagging of `\flagverse` and line numbers in verse is not ideal


mbertucci47:
Not sure what the correct tagging for `\flagverse` is, but I assume the line numbers should probably be tagged as artifacts.
```tex
\DocumentMetadata
  {
    lang=en-US,
    pdfversion=2.0,
    pdfstandard=ua-2,
    tagging=on
  }
\documentclass{article}

\usepackage{verse}

\begin{document}

\settowidth{\versewidth}{In a cavern, in a canyon,}
\begin{verse}[\versewidth]
\poemlines{2}
\begin{altverse}
\flagverse{1.} In a cavern, in a canyon, \\
Excavating for a mine, \\
Lived a miner, forty-niner, \label{vs:49} \\
And his daughter, Clementine. \\!
\end{altverse}
\begin{altverse}
\flagverse{\textsc{chorus}} Oh my darling, Oh my darling, \\
Oh my darling Clementine. \\
Thou art lost and gone forever, \\
Oh my darling Clementine \\!
\end{altverse}
\end{verse}

\end{document}
```

mbertucci47:
I must be misunderstanding how `\SuspendTagging` works but this does not work to prevent the line numbers from being tagged and I don't know why:
```tex
\DocumentMetadata
  {
    lang=en-US,
    pdfversion=2.0,
    pdfstandard=ua-2,
    tagging=on
  }
\documentclass{article}

\usepackage{verse}

\makeatletter
\renewcommand{\getmodulo@vs}{\bgroup
  \ifnum\c@modulo@vs<\@ne   % no line numbers
  \else
  \SuspendTagging{verse}%
    \ifnum\c@modulo@vs<\tw@ % every line numbered
      \vlnumfont\thepoemline
    \else
      \@tempcnta\c@poemline
      \advance\@tempcnta -\c@fvsline
      \divide\@tempcnta\c@modulo@vs
      \multiply\@tempcnta\c@modulo@vs
      \advance\@tempcnta\c@fvsline
      \ifnum\@tempcnta=\c@poemline\vlnumfont\thepoemline\fi
    \fi
  \ResumeTagging{verse}%
  \fi
\egroup}
\makeatother

\begin{document}

\settowidth{\versewidth}{In a cavern, in a canyon,}
\begin{verse}[\versewidth]
\poemlines{2}
\begin{altverse}
\flagverse{1.} In a cavern, in a canyon, \\
Excavating for a mine, \\
[...]

PR #1519, opened 2026-08-10
update tests

mbertucci47:
The marginalia test added an extra MC after a package update.

I broke the verse tests into two now that the only open issue is #1469, assuming the current tagging of poems makes sense.


FrankMittelbach:
> I broke the verse tests into two now that the only open issue is #1469, assuming the current tagging of poems makes sense.

it makes as little sense as verse in the core implementation. Neither should be lists which is inherited from 2e. Eventually both could/should change, but they aren't teribly bad this way :-)

PR #1546, opened 2026-08-22
status updates

mbertucci47:
Status updates for several packages. All hopefully self explanatory except l3draw, which I listed as compatible with the note "Not meant for document-level use. See hawkdraw for a compatible user interface to l3draw." If this is okay, I guess a similar status and note can be applied to pgf.


davidcarlisle:
I think the comment is OK, certainly at the level of the classification here. I'm sure there are documents that do use these commands directly, there are documents that do all kinds of things, but we don't need to classify all such uses.

PR #1573, opened 2026-08-31
Memoize ext again

cfr42:
Response to discussion in the `memoize` issue https://github.com/latex3/tagging-project/issues/1568. 

Please run your `l3build` workflow with the attached test file. The logs will then include a lot more information about what is happening. If `.memo`s are generated, they will include more info, too, but unfortunately github will destroy those, so hopefully the `.log` will be sufficient.

IMHO it does not make sense to add unnecessary additional packages here. Then you do not know which causes a difference when something changes. The test file here is the one I submitted earlier this year and is supposed to be more-or-less similar to the tests `latex-lab` uses for its `tikz` support.  

I cannot guess why your `l3build` aborts memoization. That's not itself an error. It is something `memoize` does in some cases. It is also something `memoize-ext` tells `memoize` to do in some cases. So it does not show anything is wrong. It is only wrong if it shouldn't have happened.

### DIFF
--- a/_data/tagging-status.yml
+++ b/_data/tagging-status.yml
@@ -9343,12 +9343,23 @@
 
 - name: memoize
   type: package
-  status: unchecked
+  status: partially-compatible
+  comments: "Tagging requires third-party package memoize-ext."
   priority: 9
   tests: false
   tasks: needs tests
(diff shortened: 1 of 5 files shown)


mbertucci47:
Looks good to me. Thanks!

PR #1582, opened 2026-09-03
More status updates

mbertucci47:
Several status updates for packages and tikz libraries. I did not copy the comment "Compatible if `alt` text given to containing `tikzpicture`" to every tikz library entry since that seemed like overkill. Let me know if opinions differ.


davidcarlisle:
I think not having the comment is fine, the alt attribute is on `tikzpicture` so really under the control of `tikz` not any particular extension lib.

davidcarlisle:
but Manfred's linter complains

```
ERROR 'tikzlibrary-cd': 'type' is 'library' but 'ctan-pkg' is missing
ERROR 'tikzlibrary-quantikz': 'type' is 'library' but 'ctan-pkg' is missing
ERROR 'tikzlibrary-quantikz2': 'type' is 'library' but 'ctan-pkg' is missing
ERROR 'tikzlibrary-zx-calculus': 'type' is 'library' but 'ctan-pkg' is missing
ERROR 4 error(s) found
Bad or missing  YAML
```

I'll fix (I really should switch that to running on commit:-)

mbertucci47:
Oh sorry I'll fix

mbertucci47:
Fixed in https://github.com/latex3/tagging-project/commit/944757ba3de285ac33a12e42baa89d07d5cb9e79

davidcarlisle:
> Oh sorry I'll fix

ok you do it then no point both of us doing it:-)

Currently the linter runs as part of the script to generate the latex version for the `check-tagging-status`  key but that's a bit late really

PR #1590, opened 2026-09-12
Afrelationship

davidcarlisle:
updating tests in a branch for potential changes to show-pdf-tags


mbertucci47:
@davidcarlisle I think this can be merged now. I reran the tests [here](https://github.com/latex3/tagging-project/actions/runs/34720348479) and they all passed except fitch which is unrelated to the new show-pdf-tags

Issue #1597, opened 2026-09-16
tikzmark library (perhaps?) partially compatible


mbertucci47:
The tikzmark library is a difficult one since the relevant `tikzpicture`s are not usually Figures in the usual sense as they are used to place arrows, braces, etc. at arbitrary points on the page, and these objects usually carry semantic meaning for the reader. I do not know what sensible tagging would look like. That being said, the base library does not error or produce structure issues, and an author could in some cases describe what is happening with Alt text.

Furthermore, the sublibraries `listings`, `ams`, and `highlighting` all have issues. The first because listings is incompatible, the second errors, and the third produces parent-child warnings.

## base library example
```tex
\DocumentMetadata{tagging=on}
\documentclass{article}
\ifdefined\Uchar
  \usepackage{lua-unicode-math}
\fi
\usepackage{tikz}
\usetikzlibrary{tikzmark,decorations.pathreplacing}

\title{tikzmark tagging test}

\begin{document}

\tikzset{tikzmark prefix=ex1-}
\[
\tikzmark{a} e^{i \pi/2} = i
\]
This\tikz[remember picture,overlay,baseline=0pt] \draw[->] (0,1em)
to[bend left] ([shift={(-1ex,1ex)}]pic cs:a); is an important
equation.

\tikzset{tikzmark prefix=ex2-}
\begin{itemize}
\item A first item,\tikzmark{b}
\item A second item,\tikzmark{c}
\item A third item.\tikzmark{d}
\end{itemize}
\begin{tikzpicture}[remember picture,overlay]
\draw[decorate,decoration={brace}] ({pic cs:c} |- {pic cs:b})
+(0,1em) -- node[right,inner sep=1em] {some items} ({pic cs:c}
|- {pic cs:d});
\end{tikzpicture}

\tikzset{tikzmark prefix=ex3-}
\begin{tikzpicture}[remember picture]
\node (a) at (0,0) {This has a \subnode{sub}{subnode} in it};
\draw[->] (0,-1) to[bend right] (sub);
\end{tikzpicture}

\tikzset{tikzmark prefix=ex4-}
\begin{tikzpicture}[remember picture,overlay]
\draw[->,line width=1mm,cyan] (pic cs:a) to[bend left] (pic cs:b);
\end{tikzpicture}
By placing the \tikzmark{a}code before the marks, the arrow goes
under the subsequent text and picture.
\begin{tikzpicture}
\filldraw[fill=gray] (0,0) circle[radius=1cm];
[...]

Issue #1485, opened 2026-07-15
fixing `fitch` package


rzach:
The `fitch` package is marked as incompatible; the test produces a PDF in which a table is nested within a list.

I've updated the code in https://github.com/OpenLogicProject/fitch like so:
https://github.com/OpenLogicProject/fitch/commit/a90efaa20186480fd3b1e7c7b79627216cc34daf
I *think* it works now. Could someone check?

u-fischer:
I don't think that the list is actually a problem (or a large problem). 

The main problem is that the math is not properly tagged. If you add 

~~~~
\DocumentMetadata
  {
    lang=en-US,
    pdfversion=2.0,
    pdfstandard=ua-2,
    tagging=on,
    tagging-setup={math/setup=mathml-SE}
  }
\documentclass{article}

  \usepackage{unicode-math}

\usepackage{fitch}
~~~~

to some fitch examples and compile with lualatex, then you get lots of warnings from math.

BTW:    The TUG conference is in Calgary: https://www.tug.org/tug2026/  and starts tomorrow with a workshop about tagging.

FrankMittelbach:
What Ulrike is saying: given your location if you have time this afternoon I m sure you could sneak in. We start at 13:00.

rzach:
I'm not sure I understand. The only messages I get look like this:
```
Inserting mathml with Hash 86504A85C2C7E5449FEC08EDE48D2DB2
====>grabbed math=macro:->\begin {nd} \hypo {1} {P\vee Q} \hypo {2} {\neg Q} \o
pen \hypo {3} {P} \have {4} {P} \r {3} \close \open \hypo {aa} {Q} \have {6} {\
neg Q} \r {2} \have {7} {\bot } \ne {aa,6} \have {8} {P} \be {7} \close \have {
9} {P} \oe {1,3-4,aa-8} \end {nd}
Inserting mathml with Hash E8EE29BAABFA2F3FFA42614AE202322A
====>grabbed math=macro:->\begin {nd} \hypo {a} {A \Rightarrow B} \by {Premise}
{} \hypo {b} {A} \by {Premise}{} \have {c} {B} \by {\ndref {a,b} (but \emph {ho
w?})}{} \have {d} {A \wedge B} \by {}{b,c} \end {nd}
Inserting mathml with Hash CC2D33D36163211DE5E9341FD3EA99DB
====>grabbed math=macro:->\nd@num {1}
[...]

PR #1598, opened 2026-09-16
status updates

mbertucci47:
Moves tikzmark to partially-compatible; see #1597.

Moves fitch to partially-compatible; see #1485.

Moves changepage to compatible after the recent package update.

Updates tests involving siunitx. If I understand correctly, there are still issues that need to be addressed so I'm leaving it at partially-compatible.


davidcarlisle:
siunitx will change again tomorrow but lets merge now anyway

PR #1599, opened 2026-09-17
update tests for siunitx change

mbertucci47:
Some of the AF's now just say `\unit` which seems odd. @josephwright Does that look right to you?


josephwright:
@mbertucci47 Fix for that coming up

josephwright:
https://github.com/josephwright/siunitx/issues/888

